PR #1019, opened 2024-03-14
IOS Username, Password, Payment and Device Linking flow

jigar-f:


jigar-f:
Sure @atavism, If you want, you can start reviewing, The Testflight build is just for Derek to see the new changes, it would be great if you could also test the auth flow.

atavism:
> The Testflight build is just for Derek to see the new changes, it would be great if you could also test the auth flow.

@jigar-f Thanks, yeah just checking if you wanted me to start looking it over! Will try testing today 👍

atavism:
I did some testing of the auth flow. I was able to successfully create an account and sign in. I did run into a few issues:

- After doing a clean reinstall of the app, and navigating to the create an account page, a temporary error occurs. Whenever I click the refresh button, I am told that the request is missing a pro token:

IMG_5568

Checking the logs, it looks like we are trying to fetch plans with a zero user ID because we never successfully create a user:

Screenshot 2024-06-10 at 5 19 36 AM

- Getting past that issue by launching the app again and creating a new user, I run into the following issue a few times creating an account:

IMG_5571

Here's what the logs show..
```
flutter: \^[[38;5;196m│ PlatformException(1, error while sign up Unexpected status code 400, null, null)
flutter: \^[[38;5;196m├┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄
flutter: \^[[38;5;196m│ #0   StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:651:7)
flutter: \^[[38;5;196m│ #1   MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:334:18)
flutter: \^[[38;5;196m│ #2   <asynchronous suspension>
flutter: \^[[38;5;196m│ #3   _CreateAccountEmailState.createAccount (package:lantern/account/auth/create_account_email.dart:150:7)
flutter: \^[[38;5;196m│ #4   <asynchronous suspension>
ERROR webclient: pro.go:391   at github.com/getlantern/lantern-client/internalsdk.(*baseModel).InvokeMethod (model.go:103)
ERROR webclient: pro.go:391   at main.proxyinternalsdk_SessionModel_InvokeMethod (go_internalsdkmain.go:602)
ERROR webclient: pro.go:391   at _cgoexp_f0a4a7ca182a_proxyinternalsdk_SessionModel_InvokeMethod (_cgo_gotypes.go:2434)
ERROR webclient: pro.go:391   at runtime.cgocallbackg1 (cgocall.go:403)
ERROR webclient: pro.go:391   at runtime.cgocallbackg (cgocall.go:322)
ERROR webclient: pro.go:391   at runtime.cgocallback (asm_arm64.s:1130)
ERROR webclient: pro.go:391   at runtime.goexit (asm_arm64.s:1222)
```

- After I successfully create an account and sign out of it, the "Sign Out" menu option disappears from the Account page, but the "Sign In" button never re-appears

IMG_5566

jigar-f:
Thanks, @atavism for testing, I think you are facing this issue because the user id is empty.

jigar-f:
@atavism I think it is ready for review, We are almost done with this, Also, it would be great, if you could test the auth flow one last time, thanks.

atavism:
While testing, I encountered this weird Replica-related error after I navigated to the account page to sign in:

Screenshot 2024-06-17 at 8 56 25 AM

jigar-f:
I think might happen when you click on the developer tab, For testing you can just make develop mode false [here](https://github.com/getlantern/lantern-client/blob/fa48e77dbfdf85d792d63714cbb705ab15eec0ef/ios/Runner/Lantern/Models/SessionModel.swift#L31)

atavism:
After I added a password to my Pro account, I was able to sign in with it successfully. 

The only issue I'm running into is the app remains Pro after I sign out instead of switching to Free:

IMG_5655 3

IMG_5656

Is that expected? Is that happening because the device is linked to my Pro account?

jigar-f:
> Is that expected? Is that happening because the device is linked to my Pro account?
Correct, Your device is still connected to the Account which is why you are seeing this, But when you signout your device is supposed to be removed. You can check logs.

atavism:
> But when you signout your device is supposed to be removed. can you check logs.

I'm not seeing the device being removed after I sign out (I checked BackOffice). I don't see any Pro API calls happening when I click the "Sign Out" button either

jigar-f:
@atavism Can you check this [code](https://github.com/getlantern/lantern-client/blob/eac51eaddb12e75300c2a87822bbeeb4bd52f852/internalsdk/session_model.go#L1614) is exists on your side. If not you might need to take the latest pull.

internalsdk/session_model.go
```diff
@@ -725,3 +1154,735 @@ func checkAdsEnabled(session *SessionModel) error {
...
+		Email:        email,
+		DeviceId:     deviceId,
+		LegacyToken:  token,
+		LegacyUserID: userId,
+	}
+
+	log.Debugf("Sign out request %+v", signoutData)
+
+	loggedOut, logoutErr := session.authClient.SignOut(context.Background(), signoutData)
+	if err != nil {
```
atavism, line 1645: I think you want to check if logoutErr is nil here instead of err?

atavism:
> @atavism Can you check this [code](https://github.com/getlantern/lantern-client/blob/eac51eaddb12e75300c2a87822bbeeb4bd52f852/internalsdk/session_model.go#L1614) is exists on your side. If not you might need to take the latest pull.

@jigar-f I pulled the latest changes, but I needed to rebuild the Go library. It's working now! Signed out successfully and the app switched to free

internalsdk/pro/pro_test.go
```diff
@@ -1,37 +1,91 @@
 package pro
 
-import (
-	"context"
-	"net/http"
-	"testing"
+// Todo
+// @Paul we might need to find way to use mock here rather calling actual api
```
atavism, line 4: Yeah we should definitely switch to using something like https://github.com/vektra/mockery

internalsdk/pro/pro.go
```diff
@@ -60,48 +91,46 @@ func NewClient(baseURL string, opts *Opts) ProClient {
 	client := &proClient{
 		userConfig: opts.UserConfig,
 	}
-	client.webclient = webclient.NewRESTClient(defaultwebclient.SendToURL(httpClient, baseURL, client.setUserHeaders(), nil))
```
atavism, line 63: I think it would be cleaner to continue setting the default headers a single time in request middleware vs. having to pass the same headers every time we call a method on RESTClient.

If the headers differ between ProClient and AuthClient, what about creating two separate clients instead? I will open a PR with some suggestions..

atavism: https://github.com/getlantern/lantern-client/pull/1101

internalsdk/pro/pro_test.go
jigar-f: @atavism What do you think about something like much simple and Morden :- https://github.com/h2non/gock

internalsdk/pro/pro.go
jigar-f: thanks, Paul, I will take a look at it shortly.

internalsdk/session_model.go
jigar-f: Thanks for caching this, Done.

internalsdk/pro/pro.go
jigar-f: So the only reason I expose the header is, In the future, we need to change headers for specific endpoints we can do this easily, without a lot of changes, Before the header was passing it tightly coupled we didn't have any flexibility on that.

internalsdk/pro/pro_test.go
atavism: @jigar-f That one looks even better. Nice find!

Crosse: We use [this](https://github.com/uber-go/mock) for mocking some things in lantern-cloud, fwiw

Crosse: ...although it seems like we have at least one other file that specifies it was created by Mockery, soooo yay for that

atavism:
I think we can merge this one now unless you want to make any additional changes, @jigar-f

atavism (approved):

jigar-f:
@atavism Let's wait for the translation, I guess we will get that in a day or two, I want to test this flow in other locals as well, As far now we tested only in English.

jigar-f:
Found the last minute issue, working on it...
cc @atavism @Derekf5

atavism:
@jigar-f Thanks! Is it this one? https://github.com/getlantern/engineering/issues/1457

jigar-f:
No, its not that, I forgot to hide some UI elements on the account management screen for other platforms.

PR #1024, opened 2024-03-18
Android Fixes

jigar-f:
Fixes issue

- https://github.com/getlantern/engineering/issues/1251#event-12105990730
- https://github.com/getlantern/engineering/issues/1250
- https://github.com/getlantern/engineering/issues/1252
- https://github.com/getlantern/engineering/issues/1172
- https://github.com/getlantern/engineering/issues/1244
- https://github.com/getlantern/engineering/issues/1164#issuecomment-2012156555


android/app/build.gradle
```diff
@@ -375,7 +375,8 @@ dependencies {
 
     implementation 'com.squareup.okhttp3:okhttp:4.9.2'
 
-    implementation "androidx.webkit:webkit:1.9.0"
+    implementation 'androidx.webkit:webkit:1.8.0'
```
atavism, line 378: Is it necessary to downgrade `androidx.webkit:webkit`?

jigar-f: Not anymore, I gonna revert this change.

jigar-f:
@atavism This is ready fo review.

android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
```diff
@@ -232,6 +254,24 @@ class InAppBilling(
     private fun handleAcknowledgedPurchases(purchases: List<Purchase>) {
         for (purchase in purchases) {
```
atavism, line 255: You could filter purchases here to only get those with a purchased state that have not yet been acknowledged:

```kotlin
purchases
.filter { it.purchaseState == Purchase.PurchaseState.PURCHASED }
.filterNot { it.isAcknowledged }
```

android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
```diff
@@ -37,10 +37,10 @@ data class ProUser(
     }
 
     val isProUser: Boolean
-        get() = isUserStatus("active")
+        get() = userLevel == "pro"
```
atavism, line 40: Should we still check for an "active" user status (which we've traditionally used to check if a user is Pro) with 
```kotlin
isUserStatus("active") || userLevel == "pro"
```
... or is it sufficient to just check for a Pro user level now?

android/app/src/main/kotlin/org/getlantern/lantern/util/PaymentsUtil.kt
```diff
@@ -25,6 +27,14 @@ import org.getlantern.mobilesdk.Logger
 
 class PaymentsUtil(private val activity: Activity) {
     private val session: LanternSessionManager = LanternApp.getSession()
+    private var paymentDialog: ProgressDialog = ProgressDialog(activity)
```
atavism, line 30: Not super important, but I think we removed use of ProgressDialog because.. 

> "ProgressDialog is a modal dialog, which prevents the user from interacting with the app. Instead of using this class, you should use a progress indicator like [ProgressBar](https://developer.android.com/reference/android/widget/ProgressBar), which can be embedded in your app's UI." 

https://developer.android.com/reference/android/app/ProgressDialog

Is it possible to show the loading spinner on the Flutter side the way we do with some other Pro calls, or is there a reason to prefer this method instead?

atavism:
> @atavism This is ready fo review.

LGTM, @jigar-f! Just had a few comments

android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
jigar-f: I asked this query to Reflog, But its better to check both, thanks.

android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
jigar-f: Ah, I am not so sure, If we do this then we lose the purchases that are not in a purchased state, or that are not acknowledged, We have to do consumeAsync regardless of the state, so the user can purchase that again

So if

The user starts Google in-app flow and then goes back without making a purchase, and if we don't do consumeAsync user will see an error 

Let me know what you think.

android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
jigar-f: Done

android/app/src/main/kotlin/org/getlantern/lantern/util/PaymentsUtil.kt
jigar-f: Thanks for letting me know, Done, Moved to the flutter side.

jigar-f:
@atavism Answered, Mind taking a look again?

android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
atavism: Thanks! We can open another PR if need be based on his feedback

android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
atavism: Ah, makes sense then. Thanks for explaining!

atavism:
@jigar-f LGTM now! Merging..

PR #1025, opened 2024-03-18
Use flutter_windows_webview on Windows

atavism:
This updates the desktop app to use https://github.com/wgh136/flutter_windows_webview instead of https://pub.dev/packages/flutter_inappwebview for the webview on Windows (flutter_inappwebview doesn't support Windows yet -- there's an initial implementation in the works though).

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -128,7 +128,7 @@ jobs:
             lfs: true
 
         # Install Flutter
-        - uses: subosito/flutter-action@v2
+        - uses: subosito/flutter-action@v2.12.0
           with:
             channel: "stable"
 
--- a/lib/plans/checkout.dart
+++ b/lib/plans/checkout.dart
@@ -150,7 +150,7 @@ class _CheckoutState extends State<Checkout>
 
   Future<void> checkProUser() async {
     final res = await ffiProUser();
-    if (!widget.isPro && res == "true") {
+    if (!widget.isPro && res.toDartString() == "true") {
       // show success dialog if user becomes Pro during browser session
       showSuccessDialog(context, widget.isPro);
     }
--- a/lib/plans/checkout.dart
+++ b/lib/plans/checkout.dart
@@ -1,4 +1,3 @@
-import 'package:flutter_windows_webview/flutter_windows_webview.dart';
 import 'package:email_validator/email_validator.dart';
 import 'package:lantern/common/common.dart';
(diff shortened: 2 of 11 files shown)


.github/workflows/build-windows.yml
```diff
@@ -128,7 +128,7 @@ jobs:
             lfs: true
 
         # Install Flutter
-        - uses: subosito/flutter-action@v2
+        - uses: subosito/flutter-action@v2.12.0
```
atavism, line 131: It's necessary to explicitly use v2.12.0 of flutter-action here since v2.13.0 is broken: https://github.com/subosito/flutter-action/issues/277#issuecomment-1974628019

jigar-f: Thanks for letting me know.

lib/plans/checkout.dart
```diff
@@ -149,6 +148,14 @@ class _CheckoutState extends State<Checkout>
     return widgets;
   }
 
+  Future<void> checkProUser() async {
+    final res = await ffiProUser();
+    if (!widget.isPro && res == "true") {
```
jigar-f, line 153: @atavism  Here I think it should be ` if (!widget.isPro && res.toDartString() == "true")`

lib/vpn/vpn_switch.dart
```diff
@@ -19,17 +19,11 @@ class _VPNSwitchState extends State<VPNSwitch> with TrayListener {
 
   @override
   void initState() {
```
jigar-f, line 21: Can we just remove this empty method

lib/vpn/vpn_switch.dart
```diff
@@ -19,17 +19,11 @@ class _VPNSwitchState extends State<VPNSwitch> with TrayListener {
...
-    if (isDesktop()) {
-      trayManager.addListener(this);
-    }
     super.initState();
   }
 
   @override
   void dispose() {
```
jigar-f, line 26: Can we just remove this empty method

lib/vpn/vpn_switch.dart
jigar-f, line 1: We are not using TrayListener? then could we remove it?

jigar-f (commented):
@atavism  Nothing big, Just added a few small comments.

lib/vpn/vpn_switch.dart
atavism: Done, thanks @jigar-f

PR #1032, opened 2024-03-28
Handle language changes over websocket channel

atavism:
When switching languages, it's no longer necessary to switch tabs and the changes show up immediately

### DIFF
--- a/lib/common/session_model.dart
+++ b/lib/common/session_model.dart
@@ -73,6 +73,13 @@ class SessionModel extends Model {
   late ValueNotifier<bool?> proxyAvailable;
   late ValueNotifier<String?> country;
 
+  // wsMessageProp parses the given json, checks if it represents a pro user message and
+  // returns the value (if any) in the map for the given property.
+  String? wsMessageProp(Map<String, dynamic> json, String field) {
+    if (json["type"] != "pro") return null;
+    return json["message"][field];
+  }
+
   Widget proUser(ValueWidgetBuilder<bool> builder) {
     if (isMobile()) {
       return subscribedSingleValueBuilder<bool>('prouser', builder: builder);
@@ -85,8 +92,7 @@ class SessionModel extends Model {
         if (websocket == null) return;
         websocket.messageStream.listen(
           (json) {
-            if (json["type"] != "pro") return;
-            final userStatus = json["message"]["userStatus"];
+            final userStatus = wsMessageProp(json, "userStatus");
             if (userStatus != null && userStatus.toString() == "active") setValue(true);
           },
           onError: (error) => appLogger.i("websocket error: ${error.description}"),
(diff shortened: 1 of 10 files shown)


lib/common/session_model.dart
```diff
@@ -183,9 +178,21 @@ class SessionModel extends Model {
...
       defaultValue: 'en',
+      onChanges: (setValue) {
+        if (websocket == null) return;
+        websocket.messageStream.listen(
+          (json) {
+            if (json["type"] != "pro") return;
+            final language = json["message"]["language"];
+            if (language != null && language != "") setValue(language);
+          },
+          onError: (error) => print(error),
```
jigar-f, line 193: Can we use appLogger here?

lib/common/session_model.dart
```diff
@@ -183,9 +178,21 @@ class SessionModel extends Model {
...
     }
+    final websocket = WebsocketImpl.instance();
     return ffiValueBuilder<String>(
       'lang',
       defaultValue: 'en',
+      onChanges: (setValue) {
+        if (websocket == null) return;
+        websocket.messageStream.listen(
+          (json) {
+            if (json["type"] != "pro") return;
```
jigar-f, line 189: Not super important, but Do you think we can create a utils method for  `if (json["type"] != "pro") return;` check we are starting to use this at multiple places, In the future if the needed change we need to change only one?

jigar-f (approved):
LGTM, Just two small comments.

jigar-f:
@atavism While you are socket changes, Could you add 

- If I go on the Pro code activation flow, If I enter code and get success but I still see an old UI, For a user to become a pro, the User has to restart the app.

lib/common/session_model.dart
atavism: Done!

atavism:
@jigar-f With this [change](https://github.com/getlantern/lantern-client/pull/1032/commits/d6c71facdf1c76013ead6f9e396c413afd6d3224), after successfully redeeming a reseller code, user data is immediately refreshed. I confirmed I get the Pro home screen

atavism:
@jigar-f Hmm, just noticed the checkout page was disabled on desktop for some reason: https://github.com/getlantern/lantern-client/blob/main/lib/plans/plan_details.dart#L131

Just committed some changes to re-enable it.

lib/common/session_model.dart
```diff
@@ -73,6 +73,13 @@ class SessionModel extends Model {
   late ValueNotifier<bool?> proxyAvailable;
   late ValueNotifier<String?> country;
 
+  // wsMessageProp parses the given json, checks if it represents a pro user message and
+  // returns the value (if any) in the map for the given property.
+  String? wsMessageProp(Map<String, dynamic> json, String field) {
```
jigar-f, line 78: Thanks for making this change.

atavism:
Thanks for the review, @jigar-f!!

PR #1033, opened 2024-04-01
Pro HTTP client updates

atavism:
Depends on https://github.com/getlantern/flashlight/pull/1371

- Remove use of `flashlight/pro/client`: https://github.com/getlantern/flashlight/compare/atavism/remove-pro-client?expand=1
- Use [go-resty](https://github.com/go-resty/resty)-based HTTP client for interacting with Pro server
- Add protobuf messages for Pro API requests and responses


jigar-f:
@atavism is this ready to test? Currently, I am working froPay changes, I think it would be best to mered this and I do changes directly on here? What do you think?

atavism:
> @atavism is this ready to test? Currently, I am working froPay changes, I think it would be best to mered this and I do changes directly on here? What do you think?

Sounds great getting this merged, thanks! I'll do some final testing today. Feel free to start reviewing it whenever you have time

jigar-f:
@atavism Cool, I start reviewing this tomorrow.

jigar-f:
@atavism have you completed your testing?

atavism:
> @atavism have you completed your testing?

@jigar-f I just finished testing desktop and fixed a couple issues (such the formatted monthly cost missing from plans for instance). The full checkout flow is working for me now (plans -> payment methods -> payment redirect)

internalsdk/pro/pro.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, errors.New("error fetching user data: %v", err)
+	}
+	return &resp, nil
+}
+
+// RedeemResellerCode redeems a reseller code for the given user
+func (c *proClient) RedeemResellerCode(ctx context.Context, req *protos.RedeemResellerCodeRequest) (*protos.BaseResponse, error) {
+	var resp protos.BaseResponse
+	if err := c.webclient.PostFormReadingJSON(ctx, "/link-code-request", req, &resp); err != nil {
```
jigar-f: Here it should be `/purchase` endpoint, I fixed it, while I was doing testing.

internalsdk/pro/responses.go
```diff
@@ -0,0 +1,37 @@
+package pro
```
jigar-f, line 1: @atavism , Do you think, we should start using protoc file from lantern-cloud?, I think that would be much better if we didn't have to maintain multiple files. Let me know what do you think?

jigar-f: FYI- we can do later also, But I think doing that we can save bit of time and we don't have keep maintain same file as server.

lib/common/session_model.dart
```diff
@@ -318,6 +318,7 @@ class SessionModel extends Model {
     // Desktop users
     Localization.locale = lang;
     final newLang = lang.toNativeUtf8();
+    Localization.locale = lang;
```
jigar-f, line 321: I think we can remove this same.

jigar-f (approved):
LGTM, Just some comments.

jigar-f:
@atavism 
- I do not see a loading bar when doing the Reseller checkout
- When I click authorize device  with Link with PIN It doesn't show the PIN

Screenshot 2024-04-18 at 5 54 05 PM

atavism:
> @atavism Just a thought, Do you think we decouple proclient? So later on we can use that for Android and IOS as well, At the moment it looks like it is tightly coupled with desktop code, Maybe can create some custom interface and pass that [NewClient](https://github.com/getlantern/lantern-client/blob/fd5252a4857ef49efa720e7e02899d136817d26c/internalsdk/pro/pro.go#L55) so we can use it for desktop and android and IOS as well.

Yep, makes sense. I am making some changes now to no longer pass settings to ProClient

lib/common/session_model.dart
atavism: Fixed. Thanks!

internalsdk/pro/responses.go
atavism: > Do you think, we should start using protoc file from lantern-cloud?, I think that would be much better if we didn't have to maintain multiple files.

Yeah, I agree it would be much better not having to maintain multiple proto files, but I don't think we can have lantern-cloud as a dependency here. One reason is because it isn't open-source and this project is now. Maybe the best solution is to move the proto files to a repo like https://github.com/getlantern/common instead and update lantern-cloud/lantern-client to use it?

internalsdk/pro/pro.go
atavism: Oh, I didn't even notice that. Thanks for fixing it!

atavism:
> Maybe can create some custom interface and pass that [NewClient](https://github.com/getlantern/lantern-client/blob/fd5252a4857ef49efa720e7e02899d136817d26c/internalsdk/pro/pro.go#L55) so we can use it for desktop and android and IOS as well.

This is done. Updating Android and iOS to use it now.

atavism:
> I do not see a loading bar when doing the Reseller checkout
When I click authorize device with Link with PIN It doesn't show the PIN

The device name was missing requesting a device linking code:

```
{"status":"error","error":"Bad input: request missing Device Name","errorId":"bad-input-missing-device-name","Code":"","ExpireAt":0}
```

Should be fixed now.

atavism:
> Maybe can create some custom interface and pass that [NewClient](https://github.com/getlantern/lantern-client/blob/fd5252a4857ef49efa720e7e02899d136817d26c/internalsdk/pro/pro.go#L55) so we can use it for desktop and android and IOS as well.

iOS is using the new Pro client now.

For Android, the changes are more extensive than I anticipated since we have to get rid of the LanternHttpClient. I opened a ticket for that.

internalsdk/pro/responses.go
jigar-f: @atavism Can't we add this [apipb](https://github.com/getlantern/lantern-cloud/blob/7372fa0b96888e2cf5cb5ae1f775a2b136850720/cmd/api/apipb/users.proto#L2) package and use it?

PR #1042, opened 2024-04-05
Replica Fixes/ Fix App caching lot of data

jigar-f:


jigar-f:
@atavism Could you please review this?

atavism:
@jigar-f Sure, taking a look!

lib/replica/home.dart
```diff
@@ -53,16 +55,38 @@ class _ReplicaHomeScreenState extends State<ReplicaHomeScreen> {
 
   @override
   Widget build(BuildContext context) {
+    print("Home called");
     // We are showing the ReplicaSearchScreen here since we want the bottom tabs to be visible (they are not if it's its own route)
     // <08-23-22, kalli>  Not ideal UX - maybe add a spinner? Not sure
     // <09-07-22, kalli> Update after testing a debug build - this is not very noticeable.
     if (showResults) {
       return ReplicaSearchScreen(
```
atavism, line 63: It looks like we use the same `textEditingController` (with `replicaSearchInput` label) in the ReplicaSearchScreen widget. Is it possible to pass the instance created here when we construct a ReplicaSearchScreen vs. maintaining two separate instances? Like this: https://github.com/getlantern/lantern-client/compare/jigar/replica-fixes...atavism/replica-fixes?expand=1

lib/replica/home.dart
```diff
@@ -53,16 +55,38 @@ class _ReplicaHomeScreenState extends State<ReplicaHomeScreen> {
...
         currentQuery: currentQuery,
         currentTab: currentTab,
+        onBackButtonPressed: onBackButtonPressed,
       );
     }
+    // we need to initialize controller again coz we when switch widget controller is getting disposed
+    _textEditingController = CustomTextEditingController(formKey: _formKey);
```
atavism, line 70: This isn't a problem with your changes, but I noticed we aren't disposing this controller when the widget is no longer needed. Should we be doing that?

```dart
@override
void dispose() {
  _textEditingController.dispose();
}    
```

lib/replica/home.dart
```diff
@@ -53,16 +55,38 @@ class _ReplicaHomeScreenState extends State<ReplicaHomeScreen> {
...
     // <08-23-22, kalli>  Not ideal UX - maybe add a spinner? Not sure
     // <09-07-22, kalli> Update after testing a debug build - this is not very noticeable.
     if (showResults) {
       return ReplicaSearchScreen(
         currentQuery: currentQuery,
         currentTab: currentTab,
+        onBackButtonPressed: onBackButtonPressed,
       );
     }
+    // we need to initialize controller again coz we when switch widget controller is getting disposed
```
atavism, line 69: Related to the comment here, I noticed without this change -- after searching for some Replica content and pressing the back button -- I run into a "CustomTextEditingController was used after being disposed" exception. Here's the logs:

```
I/flutter ( 2513): [2024-04-08 11:32:43.598937 | Catcher 2 | INFO] ---------- ERROR ----------
I/flutter ( 2513): [2024-04-08 11:32:43.599183 | Catcher 2 | INFO] setState() called after dispose(): _CTextFieldState#e6540(lifecycle state: defunct, not mounted)
I/flutter ( 2513): This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
I/flutter ( 2513): The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
I/flutter ( 2513): This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
I/flutter ( 2513): [2024-04-08 11:32:43.599240 | Catcher 2 | INFO]
I/flutter ( 2513): [2024-04-08 11:32:43.599284 | Catcher 2 | INFO] ------- STACK TRACE -------
I/flutter ( 2513): [2024-04-08 11:32:43.599403 | Catcher 2 | INFO] #0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1167:9)
I/flutter ( 2513): [2024-04-08 11:32:43.599449 | Catcher 2 | INFO] #1      State.setState (package:flutter/src/widgets/framework.dart:1202:6)
I/flutter ( 2513): [2024-04-08 11:32:43.599494 | Catcher 2 | INFO] #2      _CTextFieldState.initState.<anonymous closure> (package:lantern/common/ui/custom/text_field.dart:91:7)
I/flutter ( 2513): [2024-04-08 11:32:43.599534 | Catcher 2 | INFO] #3      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:433:24)
I/flutter ( 2513): [2024-04-08 11:32:43.599574 | Catcher 2 | INFO] #4      FocusNode._notify (package:flutter/src/widgets/focus_manager.dart:1089:5)
I/flutter ( 2513): [2024-04-08 11:32:43.599613 | Catcher 2 | INFO] #5      FocusManager.applyFocusChangesIfNeeded (package:flutter/src/widgets/focus_manager.dart:1794:12)
I/flutter ( 2513): [2024-04-08 11:32:43.599651 | Catcher 2 | INFO] #6      _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
I/flutter ( 2513): [2024-04-08 11:32:43.599695 | Catcher 2 | INFO] #7      _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
I/flutter ( 2513): [2024-04-08 11:32:43.599733 | Catcher 2 | INFO]
I/flutter ( 2513): [2024-04-08 11:32:43.599772 | Catcher 2 | INFO] ------- CUSTOM INFO -------
```

It actually looks like that's an issue with the CustomTextEditingController [here](https://github.com/getlantern/lantern-client/blob/main/lib/common/ui/custom/text_field.dart#L89)

atavism: This should fix the issue https://github.com/getlantern/lantern-client/pull/1043

atavism:
LGTM, @jigar-f! Just a couple comments, and I opened a PR here: https://github.com/getlantern/lantern-client/pull/1043

lib/replica/home.dart
jigar-f: Good Idea, Thanks for making changes.

lib/replica/home.dart
jigar-f: Thanks for caching this, I will add this.

lib/replica/home.dart
jigar-f: I actually tested changes, Somehow I am not getting this issue, But thanks for making this change.

jigar-f:
@atavism Could you test this PR again? See if you are still getting issues?

lib/replica/home.dart
jigar-f: It looks like we using two controllers we facing some issues due Life cycle of the widget. I think we should go the same for now, Also Once users start searching they will anyway goto the next screen and use that onwards.

Let me know what you think?

atavism:
LGTM!!

PR #1043, opened 2024-04-08
Replica fixes - CustomTextEditingController used after being disposed

atavism:
Fixes an issue where `setState()` is called on a TextEditingController after it has been disposed

```
I/flutter ( 6149): [2024-04-08 11:58:47.128697 | Catcher 2 | INFO] ---------- ERROR ----------
I/flutter ( 6149): [2024-04-08 11:58:47.128900 | Catcher 2 | INFO] A CustomTextEditingController was used after being disposed.
I/flutter ( 6149): Once you have called dispose() on a CustomTextEditingController, it can no longer be used.

....

I/flutter ( 2513): [2024-04-08 11:32:43.598937 | Catcher 2 | INFO] ---------- ERROR ----------
I/flutter ( 2513): [2024-04-08 11:32:43.599183 | Catcher 2 | INFO] setState() called after dispose(): _CTextFieldState#e6540(lifecycle state: defunct, not mounted)
I/flutter ( 2513): This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
I/flutter ( 2513): The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
I/flutter ( 2513): This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
I/flutter ( 2513): [2024-04-08 11:32:43.599240 | Catcher 2 | INFO]
I/flutter ( 2513): [2024-04-08 11:32:43.599284 | Catcher 2 | INFO] ------- STACK TRACE -------
I/flutter ( 2513): [2024-04-08 11:32:43.599403 | Catcher 2 | INFO] #0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1167:9)
I/flutter ( 2513): [2024-04-08 11:32:43.599449 | Catcher 2 | INFO] #1      State.setState (package:flutter/src/widgets/framework.dart:1202:6)
I/flutter ( 2513): [2024-04-08 11:32:43.599494 | Catcher 2 | INFO] #2      _CTextFieldState.initState.<anonymous closure> (package:lantern/common/ui/custom/text_field.dart:91:7)
I/flutter ( 2513): [2024-04-08 11:32:43.599534 | Catcher 2 | INFO] #3      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:433:24)
I/flutter ( 2513): [2024-04-08 11:32:43.599574 | Catcher 2 | INFO] #4      FocusNode._notify (package:flutter/src/widgets/focus_manager.dart:1089:5)
I/flutter ( 2513): [2024-04-08 11:32:43.599613 | Catcher 2 | INFO] #5      FocusManager.applyFocusChangesIfNeeded (package:flutter/src/widgets/focus_manager.dart:1794:12)
I/flutter ( 2513): [2024-04-08 11:32:43.599651 | Catcher 2 | INFO] #6      _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
I/flutter ( 2513): [2024-04-08 11:32:43.599695 | Catcher 2 | INFO] #7      _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
I/flutter ( 2513): [2024-04-08 11:32:43.599733 | Catcher 2 | INFO]
I/flutter ( 2513): [2024-04-08 11:32:43.599772 | Catcher 2 | INFO] ------- CUSTOM INFO -------
```

### DIFF
--- a/lib/common/ui/custom/text_field.dart
+++ b/lib/common/ui/custom/text_field.dart
@@ -212,19 +212,20 @@ class _CTextFieldState extends State<CTextField> {
                   end: _focusNode.hasFocus ? 2 : 0,
                 ),
                 color: white,
-                child: !_focusNode.hasFocus && widget.controller.value.text.isEmpty
-                    ? Container()
-                    : CText(
-                        widget.label,
-                        style: CTextStyle(
-                          fontSize: 12,
-                          lineHeight: 12,
-                          color: fieldKey.currentState?.mounted == true &&
-                                  fieldKey.currentState?.hasError == true
-                              ? indicatorRed
-                              : blue4,
-                        ),
-                      ),
+                child:
+                    !_focusNode.hasFocus && widget.controller.value.text.isEmpty
+                        ? Container()
+                        : CText(
+                            widget.label,
+                            style: CTextStyle(
+                              fontSize: 12,
+                              lineHeight: 12,
+                              color: fieldKey.currentState?.mounted == true &&
+                                      fieldKey.currentState?.hasError == true
+                                  ? indicatorRed
+                                  : blue4,
+                            ),
+                          ),
               )
             : widget.label,
       ],
--- a/lib/common/ui/custom/text_field.dart
+++ b/lib/common/ui/custom/text_field.dart
@@ -78,24 +78,23 @@ class CTextField extends StatefulWidget {
 }
 
 class _CTextFieldState extends State<CTextField> {
-  var hasFocus = false;
-
+  final _focusNode = FocusNode();
   final fieldKey = GlobalKey<FormFieldState>();
   final scrollController = ScrollController();
 
   @override
   void initState() {
     super.initState();
-
-    widget.controller.focusNode.addListener(() {
-      setState(() {
-        hasFocus = widget.controller.focusNode.hasFocus;
-      });
+    _focusNode.addListener(() {
+      if (!_focusNode.hasFocus) {
+        FocusScope.of(context).requestFocus(_focusNode);
+      }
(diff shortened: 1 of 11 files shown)


lib/common/ui/custom/text_field.dart
```diff
@@ -78,24 +78,23 @@ class CTextField extends StatefulWidget {
 }
 
 class _CTextFieldState extends State<CTextField> {
-  var hasFocus = false;
-
+  final _focusNode = FocusNode();
```
jigar-f, line 81: We are using _focusNode node here, can we remove it from Controller then?

jigar-f (approved):
LGTM, Just one comment.

jigar-f:
@atavism I just ran it from this branch, once I come back from replica search to the replica home I am getting this same error

WhatsApp Image 2024-04-09 at 11 37 50

jigar-f:
@atavism  After doing some debugging 

- I think we should avoid using the same controller for both widgets because when a user searches for anything and goes to ReplicaSearchScreen replicate/home. dart SearchField which is CTextField calling dispose right away so the controller is useless after that,
- Vice versa if a user comes back from search to home it will call dispose so we need to create a new controller here as well.

PR #1048, opened 2024-04-10
Update ffiPaymentRedirect to use compute

atavism:
- Update ffiPaymentRedirect to use compute
- Add PaymentRedirect API response type
- Update email exists check to return string instead of pointer

This also fixes an issue on main with desktop where the following error is shown after opening the payment redirect page

Screenshot 2024-04-10 at 3 41 32 PM

### DIFF
--- a/lib/plans/checkout.dart
+++ b/lib/plans/checkout.dart
@@ -12,7 +12,7 @@ class Checkout extends StatefulWidget {
   final Plan plan;
   final bool isPro;
 
-  Checkout({
+  const Checkout({
     required this.plan,
     required this.isPro,
     Key? key,
@@ -29,7 +29,7 @@ class _CheckoutState extends State<Checkout>
   final emailFieldKey = GlobalKey<FormState>();
   late final emailController = CustomTextEditingController(
     formKey: emailFieldKey,
-    validator: (value) => EmailValidator.validate(value ?? '')
+    validator: (value) => value!.isEmpty?null: EmailValidator.validate(value ?? '')
         ? null
         : 'please_enter_a_valid_email_address'.i18n,
   );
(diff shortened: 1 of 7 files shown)


jigar-f (approved):
LGTM!

lib/plans/checkout.dart
```diff
@@ -409,7 +406,7 @@ class _CheckoutState extends State<Checkout>
                           text: 'continue'.i18n,
                           // for Pro users renewing their accounts, we always have an e-mail address
                           // so it's unnecessary to disable the continue button
-                          disabled: !widget.isPro ? !showContinueButton : false,
+                          disabled: !enableContinueButton(),
```
jigar-f, line 409: @atavism  I was able to click the continue button without email so, I changed to use !enableContinueButton()

lib/plans/checkout.dart
```diff
@@ -69,8 +69,6 @@ class _CheckoutState extends State<Checkout>
 
   @override
   void dispose() {
-    emailController.dispose();
-    refCodeController.dispose();
```
jigar-f, line 73: We are already doing dispose on CTextField, Doing it twice was throwing error.

lib/plans/checkout.dart
```diff
@@ -29,7 +29,7 @@ class _CheckoutState extends State<Checkout>
   final emailFieldKey = GlobalKey<FormState>();
   late final emailController = CustomTextEditingController(
     formKey: emailFieldKey,
-    validator: (value) => EmailValidator.validate(value ?? '')
+    validator: (value) => value!.isEmpty?null: EmailValidator.validate(value ?? '')
```
jigar-f, line 32: It was showing an error before even entering the email, I changed it to when the user starts to enter then we threw an error.

jigar-f:
I am good with this @atavism, Merging...

lib/plans/checkout.dart
atavism: Thanks for fixing this!

atavism:
Thanks for merging and the review, @jigar-f!

PR #1049, opened 2024-04-11
Remove expectedMonthlyPrice map from Plan protobuf message

atavism:
Fix for app crash after update


atavism:
To confirm the issue is resolved, I tried installing 7.8.2 and upgrading to 7.8.4 (the app crashes), but 7.8.2 -> 7.8.5 (and 7.8.4 -> 7.8.5) works fine.

PR #1055, opened 2024-04-25
Fix issue updating user data over websocket

atavism:
In the old Pro client, whenever we fetch user data, it is stored afterwards in a local cache. Whenever that cache changes, the updates are sent over the websocket channel: https://github.com/getlantern/flashlight/blob/main/pro/user_data.go#L154

It looks this was left out of the [new Pro client](https://github.com/getlantern/lantern-client/pull/1033), and we were no longer receiving user data updates in real-time: https://github.com/getlantern/lantern-client/blob/7163452b617d7a7384d2f050bfbf80cde38db3fb/lib/common/session_model.dart#L94

### DIFF
--- a/desktop/app/pro.go
+++ b/desktop/app/pro.go
@@ -86,23 +86,41 @@ func (app *App) IsProUser(ctx context.Context) (isPro bool, ok bool) {
 	if err != nil {
 		return false, false
 	}
-	return IsProUser(ctx, app.proClient, app.settings.GetUserID())
+	return IsProUser(ctx, app.proClient, app.settings)
 }
 
-func IsProUser(ctx context.Context, proClient pro.ProClient, userId int64) (isPro bool, ok bool) {
+func IsProUser(ctx context.Context, proClient pro.ProClient, uc common.UserConfig) (isPro bool, ok bool) {
 	isActive := func(user *protos.User) bool {
 		return user != nil && user.UserStatus == "active"
 	}
-	user, found := GetUserDataFast(ctx, userId)
+	user, found := GetUserDataFast(ctx, uc.GetUserID())
 	if !found {
 		ctx := context.Background()
-		resp, err := proClient.UserData(ctx)
+		resp, err := fetchUserDataWithClient(ctx, proClient, uc)
 		if err != nil {
 			return false, false
 		}
 		user = resp.User
 	}
 	return isActive(user), true
+}
+
+func fetchUserDataWithClient(ctx context.Context, proClient pro.ProClient, uc common.UserConfig) (*pro.UserDataResponse, error) {
+	userID := uc.GetUserID()
+	log.Debugf("Fetching user status with device ID '%v', user ID '%v' and proToken %v",
+		uc.GetDeviceID(), userID, uc.GetToken())
+	resp, err := proClient.UserData(ctx)
+	if err != nil {
+		return nil, err
+	}
+	setUserData(ctx, userID, resp.User)
+	log.Debugf("User %d is '%v'", userID, resp.User.UserStatus)
+	return resp, nil
+}
+
+func setUserData(ctx context.Context, userID int64, user *protos.User) {
+	log.Debugf("Storing user data for user %v", userID)
+	userData.save(ctx, userID, user)
 }
 
 // isActive determines whether the given status is an active status
--- a/desktop/lib.go
+++ b/desktop/lib.go
@@ -258,6 +258,7 @@ func getUserData() (*protos.User, error) {
 //export hasPlanUpdatedOrBuy
 func hasPlanUpdatedOrBuy() *C.char {
 	//Get the cached user data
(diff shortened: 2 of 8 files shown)


desktop/app/pro.go
```diff
@@ -105,6 +105,24 @@ func IsProUser(ctx context.Context, proClient pro.ProClient, userId int64) (isPr
 	return isActive(user), true
 }
 
+func fetchUserDataWithClient(ctx context.Context, proClient pro.ProClient, uc common.UserConfig) (*pro.UserDataResponse, error) {
```
atavism, line 108: Reference: https://github.com/getlantern/flashlight/blob/main/pro/user_data.go#L146

desktop/app/pro.go
```diff
@@ -86,17 +86,17 @@ func (app *App) IsProUser(ctx context.Context) (isPro bool, ok bool) {
...
+	return IsProUser(ctx, app.proClient, app.settings)
 }
 
-func IsProUser(ctx context.Context, proClient pro.ProClient, userId int64) (isPro bool, ok bool) {
+func IsProUser(ctx context.Context, proClient pro.ProClient, uc common.UserConfig) (isPro bool, ok bool) {
 	isActive := func(user *protos.User) bool {
 		return user != nil && user.UserStatus == "active"
 	}
-	user, found := GetUserDataFast(ctx, userId)
+	user, found := GetUserDataFast(ctx, uc.GetUserID())
```
jigar-f, line 96: This doesn't fix the issue, If the user is on the payment page, then we already have user data, which will always return local data. Still, what we want is if a user is on the payment page then we need to call userData API again and again till the user is on the page, SO we will know then user has made the payment outside the side app, On the System browser.

desktop/lib.go
```diff
@@ -255,6 +253,29 @@ func getUserData() (*protos.User, error) {
 	return user, nil
 }
 
+// this method is reposible for checking if the user has updated plan or bought plans
+//
+//export hasPlanUpdatedOrBuy
+func hasPlanUpdatedOrBuy() *C.char {
```
jigar-f, line 259: @atavism Using this method, we check whether the user has bought plans or not, If so we show them the dialog we do in other platforms.

jigar-f: This will work for Windows as well. We don't have to rely on the user to restart the app or go back.

lib/plans/checkout.dart
```diff
@@ -473,4 +476,21 @@ class _CheckoutState extends State<Checkout>
       return false;
     }
   }
+
+
+  void hasPlansUpdateOrBuy() {
```
jigar-f, line 481: As soon as the user generates a payment link, We start calling this method and compare it with old data to check if the user has made a payment or not.

jigar-f:
@atavism Do you mind taking a look at the new changes?

desktop/lib.go
atavism: Nice, thanks! I'll try it out now

desktop/lib.go
```diff
@@ -255,6 +253,27 @@ func getUserData() (*protos.User, error) {
...
+	cahcheUserData, isOldFound := app.GetUserDataFast(context.Background(), uc.GetUserID())
+	//Get latest user data
+	resp, err := proClient.UserData(context.Background())
+	if err != nil {
+		return sendError(err)
+	}
+	if isOldFound {
+		if cahcheUserData.Expiration < resp.User.Expiration {
+			// New data has a later expiration
+			return C.CString(string("true"))
```
atavism, line 271: Nice, I think this is a simple, effective way to check if a user became Pro or renewed their subscription!

lib/plans/checkout.dart
```diff
@@ -473,4 +476,21 @@ class _CheckoutState extends State<Checkout>
...
+    try {
+      retry(() async {
```
atavism, line 483: I guess the delay between attempts is computed as `pow(2, attempt) * delayFactor` where delayFactor defaults to 200 milliseconds: https://pub.dev/documentation/retry/latest/retry/RetryOptions-class.html

"With the default configuration functions will be retried up-to 7 times (8 attempts in total), sleeping 1st, 2nd, 3rd, ..., 7th attempt:

400 ms +/- 25%
800 ms +/- 25%
1600 ms +/- 25%
3200 ms +/- 25%
6400 ms +/- 25%
12800 ms +/- 25%
25600 ms +/- 25%"

I think the max attempts is fine, but it might make sense to bump the delay factor to something like twenty seconds instead?

atavism:
@jigar-f Your changes LGTM! Just have that one comment.

desktop/lib.go
jigar-f: Thanks!

lib/plans/checkout.dart
jigar-f: I agree, I gonna make this change.

jigar-f:
@atavism added all the changes.

atavism:
LGTM!

PR #1057, opened 2024-04-29
Lantern Desktop Beta 8.0.0 collected testing issues

atavism:
Resolves https://github.com/getlantern/engineering/issues/1336

- Make text fields 56px high on desktop
- Remove floating label near cursor from text area on report issue screen

Screenshot 2024-04-28 at 9 51 20 PM

### DIFF
--- a/lib/ffi.dart
+++ b/lib/ffi.dart
@@ -21,6 +21,8 @@ void setLang(lang) => _bindings.setSelectLang(lang);
 
 String websocketAddr() => _bindings.websocketAddr().cast<Utf8>().toDartString();
 
+void ffiExit() => _bindings.exitApp();
+
 Pointer<Utf8> ffiVpnStatus() => _bindings.vpnStatus().cast<Utf8>();
 
 Pointer<Utf8> ffiSelectedTab() => _bindings.selectedTab().cast<Utf8>();
--- a/lib/ffi.dart
+++ b/lib/ffi.dart
@@ -21,7 +21,10 @@ void setLang(lang) => _bindings.setSelectLang(lang);
 
 String websocketAddr() => _bindings.websocketAddr().cast<Utf8>().toDartString();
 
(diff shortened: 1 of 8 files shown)


jigar-f (approved):
LGTM @atavism

lib/home.dart
```diff
@@ -118,7 +118,7 @@ class _HomePageState extends State<HomePage> with TrayListener, WindowListener {
   void onTrayMenuItemClick(MenuItem menuItem) async {
     switch (menuItem.key) {
       case 'show':
-        windowManager.show();
+        windowManager.focus();
```
atavism, line 121: Thanks for making this change!

atavism:
@jigar-f I made one more change to add an ffi method to exit the app (instead of calling `SystemChannels.platform.invokeMethod('SystemNavigator.pop')` that is called via Exit in the system tray menu. That seems to work much better. Do the changes still look good to you?

jigar-f:
Let me check.

jigar-f:
@atavism when exit is called by go, it's considered a crash on the flutter end. I see an error related to `Crashed Thread:        3  io.flutter.ui` I think we should use both, Go will end all go process and flutter will flutter process.

lib/ffi.dart
```diff
@@ -21,7 +21,10 @@ void setLang(lang) => _bindings.setSelectLang(lang);
 
 String websocketAddr() => _bindings.websocketAddr().cast<Utf8>().toDartString();
 
-void ffiExit() => _bindings.exitApp();
+void ffiExit() {
+  _bindings.exitApp();
+  SystemChannels.platform.invokeMethod('SystemNavigator.pop');
```
jigar-f, line 26: @atavism I think we should use both, Go will end all go process and flutter will end flutter process

atavism: Sounds good. Thanks!

PR #1064, opened 2024-05-02
Fix issues with report an issue page

atavism:
Resolves https://github.com/getlantern/engineering/issues/1361

- [x] When sending a report, it's not obvious that a description is required. Can we make the description optional?
- [x] Always show "Describe your issue" placeholder text
- [x] Do not show "send_report" label on mouse hover of Send Report button

### DIFF
--- a/lib/common/session_model.dart
+++ b/lib/common/session_model.dart
@@ -305,12 +305,16 @@ class SessionModel extends Model {
     );
   }
 
-  Future<void> setProxyAll<T>(bool on) async {
-    unawaited(
-      methodChannel.invokeMethod('setProxyAll', <String, dynamic>{
-        'on': on,
-      }),
-    );
+  Future<void> setProxyAll<T>(bool isOn) async {
+    if (isMobile()) {
+      unawaited(
+        methodChannel.invokeMethod('setProxyAll', <String, dynamic>{
+          'on': isOn,
+        }),
+      );
+      return;
+    }
+    return await compute(ffiSetProxyAll, isOn ? 'on' : 'off');
   }
 
   Future<String> getCountryCode() async {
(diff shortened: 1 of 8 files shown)


lib/common/session_model.dart
```diff
@@ -82,11 +82,17 @@ class SessionModel extends Model {
...
   late ValueNotifier<String?> country;
 
-  // wsMessageProp parses the given json, checks if it represents a pro user message and
-  // returns the value (if any) in the map for the given property.
-  String? wsMessageProp(Map<String, dynamic> json, String field) {
-    if (json["type"] != "pro") return null;
-    return json["message"][field];
+  // listenWebsocket listens for websocket messages from the server. If a message matches the given message type,
+  // the onMessage callback is triggered with the given property value
+  void listenWebsocket<T>(WebsocketImpl? websocket, String messageType, property, void Function(T?) onMessage) {
```
jigar-f, line 87: That's a neat approach, looks good! 👍

lib/common/session_model.dart
```diff
@@ -305,12 +294,16 @@ class SessionModel extends Model {
...
   }
 
-  Future<void> setProxyAll<T>(bool on) async {
-    unawaited(
-      methodChannel.invokeMethod('setProxyAll', <String, dynamic>{
-        'on': on,
-      }),
-    );
+  Future<void> setProxyAll<T>(bool isOn) async {
+    if (isMobile()) {
```
jigar-f, line 298: I think we need to remove Mobile call here, We don't have any method for proxyAll

jigar-f (approved):
LGTM!

lib/common/session_model.dart
atavism: Will do!

lib/common/session_model.dart
atavism: Thanks!

atavism:

PR #1065, opened 2024-05-03
Google in app purchase fixes

jigar-f:


android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
```diff
@@ -318,4 +331,36 @@ class InAppBilling(
...
+        json.addProperty("idempotencyKey", System.currentTimeMillis().toString())
+        json.addProperty("provider", PaymentProvider.GooglePlay.toString().lowercase())
+        json.addProperty("email", session.email())
+        json.addProperty("plan", planID)
+        json.addProperty("currency", currency.lowercase())
+        json.addProperty("deviceName", session.deviceName())
+        json.addProperty("token", token)
+
+        lanternClient.post(
+            LanternHttpClient.createProUrl("/purchase"),
```
myleshorton, line 347: I'm a little unclear on why we're adding this. Aren't we already calling the purchase endpoint in `sendPurchaseRequest` in `PaymentUtil.kt`?

jigar-f: the reason why we are adding here, if purchase takes some time process then,if server try to make acknowledgment and purchase had not been process then it will throw error, so when user restarts app, will check for any purchases that has not been acknowledged, if found we need to let server know that purchase state has changed and make acknowledgment.

android/app/src/main/kotlin/org/getlantern/lantern/model/LanternSessionManager.kt
```diff
@@ -50,7 +59,21 @@ class LanternSessionManager(application: Application) : SessionManager(applicati
...
+            Resources.getSystem().configuration.locales[0]
+        } else {
+            Resources.getSystem().configuration.locale
+        }
+        val deviceLocal = Currency.getInstance(defaultLocale).currencyCode.lowercase()
+        return if (supportedCurrencyCodes.contains(deviceLocal)) {
+            deviceLocal
+        } else {
+            "usd"
+        }
```
atavism, line 76: If the currency associated with the primary locale isn't supported, but another on the Resources.getSystem().configuration.locales list is valid, should we try using it instead of defaulting to usd?

atavism: Maybe we filter locales from Resources.getSystem().configuration.locales until we find one that is on the supported currency codes list?

android/app/src/main/kotlin/org/getlantern/lantern/model/LanternSessionManager.kt
```diff
@@ -17,6 +21,11 @@ class LanternSessionManager(application: Application) : SessionManager(applicati
 
     private var referral: String? = null
     private var verifyCode: String? = null
+    private val supportedCurrencyCodes = arrayOf(
+        "rub", "sar", "inr", "sgd", "jpy", "aud", "try", "mop", "cad", "vnd",
+        "thb", "hkd", "bhd", "eur", "php", "cny", "irr", "krw", "myr", "gbp",
+        "twd", "idr", "usd", "aed", "ils"
+    )
```
atavism, line 28: Do we know how often this list of supported currency codes is updated? I am just wondering if we end up adding a new one to the list, if we would run into issues having a bunch of old clients on an old version without it. If it changes infrequently though, I think that's fine

atavism:
LGTM! Just a couple comments. I did some quick testing to confirm we check for unacknowledged purchases whenever the app starts and the billing client is initialized..

```bash
05-06 17:43:38.419 14554 14554 D org.getlantern.mobilesdk.model.SessionManager: Configured locale was en_US, setting as default locale
05-06 17:43:38.435 14554 14554 D org.getlantern.mobilesdk.model.SessionManager: Lingver.init() finished at 56
05-06 17:43:38.476 14554 14554 D InAppBilling: Starting connection
05-06 17:43:38.790 14554 14652 D InAppBilling: onBillingSetupFinished with response code: 0
05-06 17:43:38.790 14554 14652 D InAppBilling: Updating SKUs
05-06 17:43:38.799 14554 14652 D InAppBilling: Checking for pending purchases
05-06 17:43:38.813 14554 14654 D InAppBilling: Got 0 purchases
```

android/app/src/main/kotlin/org/getlantern/lantern/model/LanternSessionManager.kt
jigar-f: @atavism  I think we should just get rid of this currency codes list, whatever we get from the device, we should just send it to the server, if the currency is not supported it will throw an error on the server and the server can update currency easily, What do you think?

android/app/src/main/kotlin/org/getlantern/lantern/model/LanternSessionManager.kt
jigar-f: I think that is a good idea, Changing...

jigar-f: Done

jigar-f:
Thanks, @atavism for review, team working on API that returns list of supported currency

android/app/src/main/kotlin/org/getlantern/lantern/model/LanternSessionManager.kt
atavism: Thanks for making that change!

jigar-f:
@atavism We decided to remove the currency list, Could you take a look at this?

android/app/src/main/kotlin/org/getlantern/lantern/util/PaymentsUtil.kt
```diff
@@ -244,12 +238,46 @@ class PaymentsUtil(private val activity: Activity) {
...
+                                session.setIsProUser(true)
+                                activity.runOnUiThread {
+                                    methodCallResult.success("purchaseSuccessful")
+                                }
+                            }
+
+                            override fun onFailure(throwable: Throwable?, error: ProError?) {
+                                Logger.error(TAG, "Unable to fetch user data: $throwable.message")
+                                activity.runOnUiThread {
+                                    methodCallResult.success("purchaseSuccessful")
```
atavism, line 260: Just checking, do we want to call `methodCallResult.success` if there's an error fetching user data?

atavism:
> @atavism We decided to remove the currency list, Could you take a look at this?

Thanks, @jigar-f .. LGTM! I just had one comment

atavism (approved):

android/app/src/main/kotlin/org/getlantern/lantern/util/PaymentsUtil.kt
jigar-f: I had the same thought on this, But then I found out that the server doesn't make sure pro is the right way and takes some time, So if the purchase is a success we should use a pro, If something is wrong we are already calling API on next time user start app, @atavism  What do you think on that?

jigar-f:
@atavism Added changes.

android/app/src/main/kotlin/org/getlantern/lantern/util/PaymentsUtil.kt
```diff
@@ -245,17 +245,18 @@ class PaymentsUtil(private val activity: Activity) {
                         * make user pro temporary next user open app it will check the purchase state and call api accordingly
                         * */
                         session.linkDevice()
+                        session.setIsProUser(true)
```
jigar-f, line 248: Added changes here,
@atavism makes user pro regress of userData API call, If payment has been processed then it pro temporary, Next user starts the app it will fetch data from server.

jigar-f:
@atavism Do not merge this yet, there is still one needs to add.

jigar-f:
@atavism Do you mind taking a look at it, this is ready to merge?

atavism:
LGTM!

PR #1071, opened 2024-05-08
Use app version set at build time on desktop

atavism:
This updates the desktop code to use the app version set at build time.

Consolidates code between desktop and mobile and removes common code that was duplicated


jigar-f (approved):
@atavism  LGTM!

atavism:

PR #1078, opened 2024-05-21
Update CI to install version of Flutter from pubspec.yaml 

atavism:
And explicitly set the version to 3.19.6 in pubspec.yaml (it's necessary to have an exact Flutter version defined for flutter-version-file to work)

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -19,7 +19,7 @@ version: 1.0.0+1
 
 environment:
   sdk: '>=3.2.3 <4.0.0'
-  flutter: '>=3.3.0'
+  flutter: 3.19.6
 
 dependencies:
   # SDKs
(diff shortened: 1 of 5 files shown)


pubspec.yaml
```diff
@@ -19,7 +19,7 @@ version: 1.0.0+1
 
 environment:
   sdk: '>=3.2.3 <4.0.0'
-  flutter: '>=3.3.0'
+  flutter: 3.19.6
```
jigar-f, line 22: @atavism That is very cool, Now we don't have to change the version everytime on CI, It should take from pubspec.

jigar-f (approved):

pubspec.yaml
atavism: Yep, totally agree. I didn't realize we could do this

PR #1086, opened 2024-05-23
System tray updates for Windows

atavism:
Resolves https://github.com/getlantern/engineering/issues/1416

Moves code for managing system tray to separate TrayHandler class and some other improvements


atavism:
I tested on Windows and confirmed it resolves the issue https://github.com/getlantern/engineering/issues/1416

lib/common/tray_handler.dart
```diff
@@ -0,0 +1,87 @@
...
+        ? 'assets/images/lantern_connected_32.ico'
+        : 'assets/images/lantern_connected_32.png';
+  }
+  return Platform.isWindows
+      ? 'assets/images/lantern_disconnected_32.ico'
+      : 'assets/images/lantern_disconnected_32.png';
+}
+
+
+class TrayHandler with TrayListener {
```
jigar-f, line 20: I like having a separate class for this, thanks for making changes.

lib/common/tray_handler.dart
```diff
@@ -0,0 +1,87 @@
...
+      ],
+    );
+    await trayManager.setContextMenu(menu);
+  }
+
+  @override
+  void onTrayMenuItemClick(MenuItem menuItem) {
+    switch (menuItem.key) {
+      case 'show':
+        windowManager.show();
```
jigar-f, line 59: @atavism I think we should use `windowManager.focus();` instead of  `windowManager.show();`

jigar-f (commented):
@atavism Just a couple of comments otherwise LGTM

lib/common/tray_handler.dart
atavism: Not sure I understand the difference (show is supposed to show and give focus on the window) but I can change it

jigar-f: So if you have multiple windows open on Mac then if you click to show it does not show up foreground, I think we talked about this before, when I working on some desktop fixes, that's why I remember,

jigar-f:
@atavism Looks good on macOS.

atavism:
> Looks good on macOS.

Cool, thanks! Merging then..

PR #1087, opened 2024-05-28
Desktop notification updates

atavism:
Resolves https://github.com/getlantern/engineering/issues/1379, depends on https://github.com/getlantern/notifier/pull/8

Updates to use new notifier that works with macOS Sonoma and to use newer version of notifu on Windows


jigar-f:
@atavism Does the notification work with this version?

atavism:
@jigar-f Yes, I tested both macOS and Windows (notifier doesn't currently work with Linux)

PR #1088, opened 2024-05-31
Add hit_lc_proxy.bash script

atavism:


desktop/lib.go
```diff
@@ -62,14 +62,18 @@ var issueMap = map[string]string{
 }
 
 //export start
-func start() {
+func start(cfgDir, proxyAll *C.char) {
```
hwh33, line 65: I don't think this is a sustainable model for passing configuration from the command-line to flashlight...

There are [many flags](https://github.com/getlantern/flashlight/blob/0aa683aeb620fd3d6e318cf788347b7b9c03429f/flags.go#L14-L44) flashlight reads in.  Some of these we may not use often, but looking through the list, I see at least 7 which we use regularly.  This also complicates the process of adding new flags.

However, you can pass arguments to flashlight through environment variables, prefixed with `LANTERN_` like `LANTERN_HEADLESS=true`.  Will that work here?  Will flutter pass environment variables through so that they can be read by `flashlight.ParseFlags`?  If so, `hit_lc_proxy.bash` could define `LANTERN_CONFIGDIR` and `LANTERN_PROXYALL`.  Conveniently, users of the script could define flags as needed via environment variables as well.

I tried to test this myself, but I can't get the script to work.

hit_lc_proxy.bash
hwh33, line 1: We can just call this `hit_proxy.bash` now.  The `lc` part of the name is a holdover from when we needed to support both old and new style proxies.  If we ever need to hit old-style proxies from Lantern Client, we could define a `hit_old_proxy.bash`.

hit_lc_proxy.bash
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# This script allows running against a specific lantern-cloud proxy identified by the PFE IP
+# It assumes that lantern-cloud is located on the filesystem as a sibling of lantern-desktop
```
hwh33, line 4: I think this assumption may no longer be valid (I don't know, maybe everyone else has all of their Lantern repos in the same folder?  I only have the Go ones in the same folder).

We could add a variable for this:
```
# Assume lantern-cloud is a sibling of lantern-client unless told otherwise.
LANTERN_CLOUD=$([[ ! -z "$LANTERN_CLOUD" ]] && echo "$LANTERN_CLOUD" || echo "../lantern-cloud")
```

then use it below, like:
```
$LANTERN_CLOUD/bin/ptool ...
```

We would need to get rid of the `-u` flag on `set` (line 6), but I'm not sure why we'd need that anyway.  This is a pretty simple script and bugs due to unset variables should be easy to trace.

hit_lc_proxy.bash
```diff
@@ -0,0 +1,18 @@
...
+
+PROXY="${1:?please specify the proxy IP}"
+TMPDIR="${TMP:-/tmp}/hit_lc_proxy/$PROXY"
+OUTFILE="$TMPDIR/proxies.yaml"
+
+rm -rf "$TMPDIR"
+mkdir -p "$TMPDIR"
+
+echo "Generating config for ${PROXY} in ${OUTFILE}..."
+../lantern-cloud/bin/ptool route dump-config --legacy "$PROXY" > "$OUTFILE"
```
hwh33, line 16: Just as a heads up, `ptool` is [going away soon](https://github.com/getlantern/lantern-cloud/pull/682), to be subsumed by `lantern-cloud/bin/lc`.  I'll try to keep track of this and update as necessary.

hwh33 (commented):
Thanks for jumping on this so quickly!

Unfortunately I can't get the script to run.  I get the following error:
```
➜  lantern-client git:(atavism/hit-lc-proxy) ✗ ./hit_lc_proxy.bash 02f6dd38-229e-4ffa-a1bc-7ad0ac74f13c 2>&1 | tee log.txt
Generating config for 02f6dd38-229e-4ffa-a1bc-7ad0ac74f13c in /tmp/hit_lc_proxy/02f6dd38-229e-4ffa-a1bc-7ad0ac74f13c/proxies.yaml...
Launching lib/main.dart on macOS in debug mode...
Building macOS application...
lib/ffi.dart:201:18: Error: Too many positional arguments: 0 allowed, but 2 found.
Try removing the extra positional arguments.
  _bindings.start(configDir.toPointerChar(), proxyAll.toString().toPointerChar());
                 ^
Target kernel_snapshot failed: Exception

Command PhaseScriptExecution failed with a nonzero exit code
/Users/harryharpham/Code/Lantern/lantern-client/macos/Pods/Pods.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.11, but the range of supported deployment target versions is 10.13 to 14.5.99. (in target 'OrderedSet' from project 'Pods')
** BUILD FAILED **

Error: Build process failed
```

Also this seems to break `flutter run -d macos` with the same error?  I think my suggestion to use environment variables instead would fix all of this (see my comment on desktop/lib.go).

atavism:
> Unfortunately I can't get the script to run. I get the following error:

You need to run `make darwin ffigen` first.

desktop/lib.go
atavism: > However, you can pass arguments to flashlight through environment variables, prefixed with LANTERN_ like LANTERN_HEADLESS=true

@hwh33 Yeah, I initially asked on Slack if we wanted to just set these arguments using `.env`:  "Right now, we are using a .env to configure the application so I think that would be the best place to add support for it? So if a CONFIG_DIR is present there, then we  override the config directory used [here](https://github.com/getlantern/lantern-client/blob/main/desktop/lib.go#L647) with it. On the Go side, we could just use https://github.com/joho/godotenv"

hit_lc_proxy.bash
atavism: @hwh33 Sounds good! Just renamed it

hit_lc_proxy.bash
atavism: Thanks.. I basically just copied the script from lantern-desktop and didn't look at the comment. I do have all my Lantern repos in the same folder but I made the change you suggested to use a variable for lantern-cloud..

hit_lc_proxy.bash
atavism: Thanks for the heads up! Didn't know that. I can switch to `lantern-cloud/bin/lc` once it's fully deprecated

desktop/lib.go
atavism: >  If so, hit_lc_proxy.bash could define LANTERN_CONFIGDIR and LANTERN_PROXYALL

Just made that change to the script, and I confirmed it works.

```bash
LANTERN_CONFIGDIR=$TMPDIR LANTERN_PROXYALL=true flutter run -d macOS
```

hwh33: > Yep, I initially asked on Slack if we wanted to use a .env to pass the args with environment variables.. Let me make that change

Oh sorry for over-explaining then!  I didn't realize that was what you meant.

> Just made that change to the script, and I confirmed it works.

Awesome, thanks!

hit_lc_proxy.bash
hwh33: Oh yeah no worries, I figured that was the case.  Cool, thanks!

hwh33:
Awesome @atavism!  This all looks great.

Unfortunately, I can't get it to respect the proxy I'm trying to specify.  I'm not sure why though.  I think we need `LANTERN_STICKYCONFIG` and maybe `LANTERN_READABLECONFIG`, but even with those it doesn't work and everything else looks right 🤔

I'll dig in more next week!

PR #1093, opened 2024-06-11
Remove duplicate strings from en.po and pull in latest translations

atavism:


jigar-f (approved):
@atavism  Changes are LGTM but did you confirm this works, I remember we had to change asset files names due to localization issue?

atavism:
@jigar-f Yes, I confirmed it works.

PR #1097, opened 2024-06-12
Remove work flow temporally

jigar-f:


jigar-f:
@atavism @jay-418 Is there anything else I need to comment out, I don't want to put on a new build or something while I am testing for ios.

atavism:
> I don't want to put on a new build or something while I am testing for ios.

This should suffice.. There's no way to release a build to beta or production anyway without creating and pushing a tag with the version number.

PR #1100, opened 2024-06-14
updating flashlight version

WendelHime:
- [ ] New code should not be using FutureBuilder (use ValueListenableBuilder instead)
- [ ] All strings are defined using localized message keys like `'my_message_key'.i18n` and are defined in `en.po`.
- [ ] All colors are defined using Hex (not using built-in colors), are defined in `colors.dart` and are not duplicated
- [ ] All text styles are defined in `text_styles.dart` and are not duplicated
- [ ] All icons are using the vector resource from Figma (do not use built-in Icons)
- [ ] Repeated code has been factored into custom widgets
- [ ] Layout looks good in both LTR (English) and RTL (Persian) languages
- [ ] If you refactored existing code, have you tested the refactored functionality against the old version to make sure you didn't break anything?
- [ ] Do the tests pass? Consistently?
- [ ] Did this change improve test coverage?
- [ ] Is the code in question being linted? If not, consider adding a linter step to CI. If yes, make sure the linter is happy.
- [ ] Have you logged tickets for related technical debt with the label “techdebt”?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/getlantern/lantern-client
 
-go 1.22.3
+go 1.22.0
 
 // replace github.com/getlantern/flashlight/v7 => ../flashlight
 
--- a/go.mod
+++ b/go.mod
@@ -22,8 +22,6 @@ replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110
 
 replace github.com/keighl/mandrill => github.com/getlantern/mandrill v0.0.0-20221004112352-e7c04248adcb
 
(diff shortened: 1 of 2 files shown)


hwh33 (approved):
LGTM!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/getlantern/lantern-client
 
-go 1.22.0
+go 1.22.3
```
garmr-ulfr, line 3: Would you revert this back to just 1.22? We won't be updating the go minor version in `go.mod`, as discussed in stand-up.

garmr-ulfr (approved):
Other than reverting the go version, this looks good to me. You can go ahead an merge after updating the version.

PR #1101, opened 2024-06-18
iOS Username and Password: Code Review Suggestions

atavism:
- Continue using request middleware to set default headers
- Move auth related code to separate package
- Do not set Content-Type header since go-resty auto-detects it


internalsdk/pro/pro_test.go
```diff
@@ -1,91 +0,0 @@
...
-// 		Salt:     []byte("salt"),
-// 		Verifier: []byte("verifier"),
-// 	}
-// 	res, e := client.SignUp(context.Background(), prepareRequestBody)
-// 	if !assert.NoError(t, e) {
-// 		return
-// 	}
-// 	log.Debugf("Got response: %v", res)
-// 	assert.NotNil(t, res)
-// }
```
atavism, line 91: I just removed this since we commented out all the code anyway

internalsdk/pro/pro.go
```diff
@@ -382,153 +352,9 @@ func (c *proClient) UserLinkValidate(ctx context.Context, code string) (*UserRec
...
-		return nil, err
-	}
-	return &resp, nil
-}
-
-// Sign up API
-// SignUp is used to sign up a new user with the SignupRequest
-func (c *proClient) SignUp(ctx context.Context, signupData *protos.SignupRequest) (bool, error) {
-	var resp protos.EmptyResponse
-	err := c.webclient.PostPROTOC(ctx, "/users/signup", nil, signupData, &resp, c.defaultHeader())
```
atavism, line 409: Did we need to use the default headers for the `/users/signup` endpoint or was that a typo?

internalsdk/auth/auth.go
```diff
@@ -0,0 +1,214 @@
...
+
+// NewClient creates a new instance of AuthClient
+func NewClient(baseURL string, opts *webclient.Opts) AuthClient {
+	httpClient := opts.HttpClient
+	if httpClient == nil {
+		httpClient = &http.Client{}
+	}
+	webclient := webclient.NewRESTClient(defaultwebclient.SendToURL(httpClient, baseURL, func(client *resty.Client, req *resty.Request) error {
+		req.SetHeader(common.ContentType, "application/x-protobuf")
+		if req.RawRequest.URL != nil && strings.HasPrefix(req.RawRequest.URL.Path, "/users/salt") {
```
atavism, line 56: I noticed we weren't setting any headers when calling the `/users/salt` endpoint. If that was unintentional, you can just remove these lines

internalsdk/auth/auth.go
```diff
@@ -0,0 +1,214 @@
...
+	}, &resp)
+	if err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
+
+// Sign up API
+// SignUp is used to sign up a new user with the SignupRequest
+func (c *authClient) SignUp(ctx context.Context, signupData *protos.SignupRequest) (bool, error) {
```
jigar-f, line 87: on signUp, We still need to pass all default headers as we were passing before.

jigar-f (approved):
LGTM!

jigar-f:
The tiny problem, I see is that if in the future we want to pass a custom header then it will become cumbersome since we don't have the option to pass a custom header to the endpoint, We will end up adding passing a lot of ifs in auth client. That I choose to pass the header in each endpoint.

internalsdk/auth/auth.go
atavism: Ah ok, thanks!

atavism:
> tiny problem, I see is that if in the future we want to pass a custom header then it will become cumbersome since we don't have the option to pass a custom header to the endpoint, We will end up adding passing a lot of ifs in auth client. That why I choose to pass the header in each endpoint.

Are we passing custom headers to any specific endpoints right now? I think there's like a single place where we aren't setting any headers calling the `/users/salt` endpoint: https://github.com/getlantern/lantern-client/blob/jigar/username-password/internalsdk/pro/pro.go#L398 (is that actually necessary?).

atavism:
> @atavism Just more thing off topic but, Suddely now we [statsTrackr](https://github.com/getlantern/lantern-client/blob/408ae8582cbc99358f1224c019fbcf8c2bb6c6d1/internalsdk/ios/ios.go#L222) is not getting triggered at all, even tracker.AddListener does not get callback, Any idea whats happening?

Checking..

internalsdk/auth/auth.go
atavism: I see `/signup` is the only `/users` endpoint where we require some of the default Pro headers https://github.com/getlantern/lantern-cloud/blob/main/cmd/api/users/mount.go#L45

atavism:
Ok, the stats tracker issue is resolved and I tested all the auth methods are working. Merging

PR #1107, opened 2024-06-24
Update pro client request middleware

atavism:
This PR updates defaultwebclient to use a different type of request middleware where RawRequest is not nil.

go-resty has "three types of request middleware, and they are performed in sequence:

- udBeforeRequest chain (registered by Client.onBeforeRequest)
- beforeRequest chain (default middleware registered by resty)
- preReqHook hook (registered by Client.SetPreRequestHook)

RawRequest is initialized at the second chain. Therefore, we get nil in our onBeforeRequest function."

### DIFF
--- a/internalsdk/pro/pro.go
+++ b/internalsdk/pro/pro.go
@@ -63,6 +63,7 @@ func NewClient(baseURL string, opts *Opts) ProClient {
 	return client
 }
 
+// prepareProRequest normalizes requests to the pro server with device ID, user ID, etc set.
 func prepareProRequest(userConfig func() common.UserConfig) func(client *resty.Client, req *http.Request) error {
 	return func(client *resty.Client, req *http.Request) error {
 		uc := userConfig()
--- a/internalsdk/pro/pro.go
+++ b/internalsdk/pro/pro.go
@@ -59,13 +59,13 @@ func NewClient(baseURL string, opts *Opts) ProClient {
 	client := &proClient{
 		userConfig: opts.UserConfig,
 	}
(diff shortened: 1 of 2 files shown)


jigar-f:
Reviewing.

internalsdk/pro/pro.go
```diff
@@ -60,40 +59,21 @@ func NewClient(baseURL string, opts *Opts) ProClient {
...
-func (c *proClient) setUserHeaders() func(client *resty.Client, req *resty.Request) error {
-	return func(client *resty.Client, req *resty.Request) error {
-
-		uc := c.userConfig()
-
+// prepareProRequest normalizes requests to the pro server with device ID, user ID, etc set.
+func prepareProRequest(userConfig func() common.UserConfig) func(client *resty.Client, req *http.Request) error {
+	return func(client *resty.Client, req *http.Request) error {
+		uc := userConfig()
 		req.Header.Set("Referer", "http://localhost:37457/")
```
jigar-f, line 70: @atavism Why do we need to pass Referer header, Can we remove it?

jigar-f (approved):
LGTM!, Just one comment.

internalsdk/pro/pro.go
atavism: I think that was originally added because a particular Pro server endpoint expected a 'Referer' header to be passed to where to redirect the user based on the app's success/fail states. I think it should be fine to omit it now as that's no longer required on the pro server

atavism:

PR #1114, opened 2024-07-02
Add logging around SQLiteException executing transactions

atavism:
Depends on https://github.com/getlantern/db-android/pull/16

Hopefully this resolves or at least gives us some more insight into the "cannot start a transaction within a transaction" exception: https://github.com/getlantern/engineering/issues/1244

### DIFF
--- a/android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
+++ b/android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
@@ -7,14 +7,14 @@ import org.joda.time.Months
 data class ProUser(
     val userId: Long,
     val token: String,
-    val referral: String,
-    val email: String,
-    val userStatus: String,
-    val code: String,
-    val subscription: String,
-    val expiration: Long,
-    val devices: List<Device>,
-    val userLevel: String,
+    val referral: String = "",
+    val email: String = "",
+    val userStatus: String = "",
+    val code: String = "",
+    val subscription: String = "",
+    val expiration: Long = 0,
+    val devices: List<Device> = mutableListOf<Device>(),
+    val userLevel: String = "",
 ) {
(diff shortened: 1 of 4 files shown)


android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
```diff
@@ -7,14 +7,14 @@ import org.joda.time.Months
...
-    val devices: List<Device>,
-    val userLevel: String,
+    val referral: String = "",
+    val email: String = "",
+    val userStatus: String = "",
+    val code: String = "",
+    val subscription: String = "",
+    val expiration: Long = 0,
+    val devices: List<Device> = mutableListOf<Device>(),
+    val userLevel: String = "",
```
atavism, line 17: It looks like the app isn't crashing anymore without these changes, but I think setting default values here is helpful to avoid the issue in the future

jigar-f: I agree with you 100%, Most values should be nullable or have default values to avoid crashes.

jigar-f:
@atavism  Changes look good, But it seems like build is failing on [jitpack.io](https://jitpack.io/#getlantern/db-android/cff7cfa7e5)

Screenshot 2024-07-03 at 11 01 26 AM

android/app/src/main/kotlin/org/getlantern/lantern/model/ProUser.kt
atavism: Let's just put out a new release with these changes then.

jigar-f: @atavism Can we wait till tomorrow? I think I might caught bug,

jigar-f:
This change has been added to #1116

PR #1117, opened 2024-07-04
Add tryParseJson utility method

atavism:


android/app/src/main/kotlin/org/getlantern/lantern/util/LanternHttpClient.kt
```diff
@@ -91,27 +93,8 @@ open class LanternHttpClient : HttpClient() {
...
-                            val userLevel = result.get("userLevel").asString
-                            val user = ProUser(
-                                userId, token, referral, email, userStatus, "", "en_us", "", 0L,
-                                mutableListOf(), userLevel
-                            )
-                            cb.onSuccess(response!!, user)
-
-                        }
-
+                        var user: ProUser? = JsonUtil.tryParseJson<ProUser?>(result.toString())
```
atavism, line 96: You can test this with the following:

```kotlin
var user: ProUser? = JsonUtil.tryParseJson<ProUser?>("""{"userId":372788064,"code":"8YXDWM","token":"BBsWlZSW3-GVm0eQJHtmegFv4IYW77EaeBmFvhMtPJcF-vZc549hgw","referral":"8YXDWM","locale":"en_US","servers":[],"bonusDays":"0","bonusMonths":"0","inviters":[],"invitees":[],"devices":null,"yinbiEnabled":false}""")
user?.let { cb.onSuccess(response!!, it) }
```

android/app/src/main/kotlin/org/getlantern/lantern/util/LanternHttpClient.kt
```diff
@@ -65,8 +65,10 @@ open class LanternHttpClient : HttpClient() {
 
             override fun onSuccess(response: Response?, result: JsonObject?) {
                 result?.let {
-                    val user: ProUser = JsonUtil.fromJson<ProUser>(result.toString())
-                    cb.onSuccess(response!!, user)
+                    val user: ProUser? = JsonUtil.tryParseJson<ProUser?>(result.toString())
```
jigar-f, line 68: @atavism The only issue is here, while parsing JSON, if we get into any issue,  I guess it will never return a response and even if it returns a response then it will return a null guess.

With an older approach, if we get into parsing issue, It will still parse some basic values and return results, so app does not crash.

jigar-f (commented):
@atavism Added one comment, otherwise LGTM

android/app/src/main/kotlin/org/getlantern/lantern/util/LanternHttpClient.kt
atavism: My thinking with this change is if the API suddenly behaves in some unexpected manner (returns null for some field when it previously always returned an empty array), we should treat that response as invalid and skip over it vs. trying to extract valid fields from it. With either approach, the app shouldn't crash now

PR #1118, opened 2024-07-08
Remove use of USE_FULL_SCREEN_INTENT permission

atavism:
The latest build submitted to the Play store was rejected because they found "that [our] app is not compliant with how the USE_FULL_SCREEN_INTENT permission is allowed to be used. Specifically, the use of the permission is not directly related to the core purpose of the app ... Apps will only be automatically granted to use the USE_FULL_SCREEN_INTENT permission if the core functionality of their app falls under one of the below categories that require high priority notifications: setting an alarm or receiving phone or video calls"


jigar-f (approved):
LGTM!

atavism:
Thanks, @jigar-f!

PR #1120, opened 2024-07-10
Remove commented permssion.

jigar-f:
Somehow even with commented permission, the Play store rejected the app.


jigar-f:
@atavism once you approve, Do upload new build.

atavism:
@jigar-f They rejected the latest submission because a previous version code (464013038) uses the USE_FULL_SCREEN_INTENT permission. So I think we need to replace that and these changes aren't necessary

PR #1127, opened 2024-07-16
[docs] Add dependency for Linux

Jovis7:
* libayatana-appindicator (required by tray_manager)
* gstreamer (required by flutter plugin audioplayers)
* libclang-dev (required by ffigen)


atavism:
Thanks for adding this to the README, @Jovis7!

PR #1128, opened 2024-07-19
Check Google Play service is available, if not fallback 

jigar-f:


android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
```diff
@@ -54,6 +56,15 @@ class InAppBilling(
     @Volatile
     private var purchasesUpdated: PurchasesUpdatedListener? = null
 
+
+    fun isPlayStoreAvailable(): Boolean {
+        if (billingResult != null) {
```
atavism, line 61: Beyond checking the billing client is initialized, does it make sense to use something like [GoogleApiAvailability.isGooglePlayServicesAvailable()](https://developers.google.com/android/reference/com/google/android/gms/common/GoogleApiAvailability#isGooglePlayServicesAvailable%28android.content.Context%29) here to check that Google Play services is installed and enabled on the device?

atavism:
LGTM, just one comment

android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
jigar-f: According to this-:https://stackoverflow.com/questions/52858403/check-if-in-app-billing-via-playstore-is-available-to-user, billingResult will work just fine.

desktop/lib.go
```diff
@@ -85,7 +85,7 @@ func start() {
 	settings := loadSettings(cdir)
 	webclientOpts := &webclient.Opts{
 		HttpClient: &http.Client{
-			Transport: proxied.ParallelPreferChained(),
+			Transport: proxied.ChainedThenFronted(),
```
jigar-f, line 88: @atavism I had changed this due to an error, You can check Slack more about that, For now even using proxied.ChainedThenFronted gives error on the first request.

jigar-f:
@atavism is all good here?, Should I go ahead and merge this?

atavism:
@jigar-f Yep, this one LGTM. Merging...

android/app/src/main/kotlin/org/getlantern/lantern/model/InAppBilling.kt
atavism: Sounds good!

PR #1134, opened 2024-07-25
Use auth feature flag

atavism:
Only show auth-related components in the app if the auth feature is enabled.

Until we have confirmation of what the auth flow looks like on iOS, this gives us the ability to disable username/password on other platforms.

### DIFF
--- a/desktop/app/app.go
+++ b/desktop/app/app.go
@@ -284,13 +284,10 @@ func (app *App) checkEnabledFeatures() {
 
 // IsFeatureEnabled checks whether or not the given feature is enabled by flashlight
 func (app *App) IsFeatureEnabled(feature string) bool {
-	enabledFeatures := app.flashlight.EnabledFeatures()
-	for enabledFeature, isEnabled := range enabledFeatures {
-		if isEnabled && enabledFeature == feature {
-			return true
-		}
+	if app.flashlight == nil {
+		return false
 	}
-	return false
+	return app.flashlight.EnabledFeatures()[feature]
 }
 
 // startFeaturesService starts a new features service that dispatches features to any relevant listeners.
--- a/desktop/app/app.go
+++ b/desktop/app/app.go
@@ -39,7 +39,6 @@ import (
 
 	"github.com/getlantern/lantern-client/desktop/autoupdate"
 	"github.com/getlantern/lantern-client/desktop/datacap"
-	"github.com/getlantern/lantern-client/desktop/features"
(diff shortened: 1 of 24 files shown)


internalsdk/android.go
```diff
@@ -616,6 +622,10 @@ func run(configDir, locale string, settings Settings, session PanickingSession)
 		log.Debugf("Chat enabled? %v", chatEnabled)
 		session.SetChatEnabled(chatEnabled)
 
+		authEnabled := runner.FeatureEnabled("auth", common.ApplicationVersion)
```
jigar-f, line 625: @atavism Did we add this flag on flashlight?

desktop/app/app.go
```diff
@@ -275,28 +275,24 @@ func (app *App) Run(isMain bool) {
...
 	enabledFeatures := app.flashlight.EnabledFeatures()
-
-	app.setFeatures(enabledFeatures, features.EnabledFeatures)
-
 	log.Debugf("Starting enabled features: %v", enabledFeatures)
 	//go app.startReplicaIfNecessary(enabledFeatures)
 }
 
+// IsFeatureEnabled checks whether or not the given feature is enabled by flashlight
+func (app *App) IsFeatureEnabled(feature string) bool {
```
jigar-f, line 286: @atavism Since this is a map, Can we do something like this?

```
func (app *App) IsFeatureEnabled(feature string) bool {
	enabledFeatures := app.flashlight.EnabledFeatures()[feature]
	return enabledFeatures
}
```

internalsdk/android.go
jigar-f: Did we add this to [global.ymal.tmpl](https://github.com/getlantern/flashlight/blob/a736060c32caf80ad772507c1fb762547489d8fe/embeddedconfig/global.yaml.tmpl#L1)

jigar-f (approved):
Just one comment.

jigar-f:
@atavism I have added some more changes, To hide username password changes.

internalsdk/android.go
atavism: @jigar-f We had an "auth" feature flag from before specifically to add username/password support that's no longer being used. I'm just re-using that

desktop/app/app.go
atavism: Oh, definitely! Thanks. Just made that change.

atavism:
@jigar-f, I made a change so that you can enable auth without having to mess with feature flagging as well. Just set `ENABLE_AUTH_FEATURE=true` in the .env

PR #1138, opened 2024-07-31
Update product IDS

jigar-f:
Added support for 1 Month plans for Google in-app purchase

### DIFF
--- a/android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
+++ b/android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
@@ -170,6 +170,7 @@ class SessionModel(
                     )
                 }
                 catch (e: Exception) {
+                    Logger.error(TAG, "Error making purchase", e)
                     result.error(
                         "unknownError",
                         activity.resources.getString(R.string.error_making_purchase),
--- a/android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
+++ b/android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
@@ -161,11 +161,23 @@ class SessionModel(
             }
 
 
-            "submitGooglePlayPayment" -> paymentsUtil.submitGooglePlayPayment(
-                call.argument("email")!!,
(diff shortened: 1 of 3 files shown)


jigar-f:
@atavism I added a new product-Id in the play console, It still not reflecting on the app, Please check after a few hours if working go ahead and test all payment flow, if all working we need to put out a new release for this.

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
```diff
@@ -161,11 +161,23 @@ class SessionModel(
...
+            "submitGooglePlayPayment" -> {
+                try {
+                    paymentsUtil.submitGooglePlayPayment(
+                        call.argument("email")!!,
+                        call.argument("planID")!!,
+                        result,
+                    )
+                }
+                catch (e: Exception) {
+                    result.error(
```
atavism, line 173: Do you want to log the exception thrown here?

atavism (approved):
LGTM! Just one comment

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
jigar-f: We do, so if something goes wrong, it should send the error to Flutter and update the UI according. Before it was just loading forever

jigar-f:
Let's wait till new changes reflect, then we can merge,

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
atavism: I thought the error being shown is R.string.error_making_purchase and not the actual message from the exception

jigar-f: @atavism Added log exception.

jigar-f:
@atavism I can see a new plan on google play. I think we can go ahead and merge this.

atavism:
@jigar-f Sounds great. Merging!

PR #1150, opened 2024-08-16
Updates to how we subscribe to websocket changes

atavism:


atavism:
@jigar-f This one is still a work-in-progress, but feel free to take it for a spin and let me know how it works for you  if you get a chance.

jigar-f:
@atavism Thanks, I'll take it for a spin next week.

atavism:
@jigar-f Thanks!! This one is working really well for me (there's a couple areas I haven't tested yet like payment redirect and reporting an issue) so I'm curious how it works for you

desktop/lib.go
```diff
@@ -76,6 +74,24 @@ var issueMap = map[string]string{
 	"Other":                       "9",
 }
 
+/*func initLantern() *app.App {
+		flags         := flashlight.ParseFlags()
```
jigar-f, line 78: @atavism Can we remove this commented code?

lib/common/session_model.dart
```diff
@@ -63,16 +71,20 @@ class SessionModel extends Model {
...
     } else {
-      country = ffiValueNotifier(ffiLang, 'lang', 'US');
-      proxyAvailable = ffiValueNotifier(
-        ffiSucceedingProxy,
-        'hasSucceedingProxy',
-        false,
-      );
-      userEmail = ffiValueNotifier(ffiEmailAddress, 'emailAddress', "");
-      proUserNotifier = ffiValueNotifier(ffiProUser, 'prouser', false);
+      configNotifier = ValueNotifier<ConfigOptions?>(null);
```
jigar-f, line 74: So now, we are updating notifier values from the WebSocket class

lib/common/websocket_subscriber.dart
```diff
@@ -0,0 +1,99 @@
+import 'common.dart';
+import 'common_desktop.dart';
+import 'package:collection/collection.dart';
+import 'package:lantern/plans/utils.dart';
+import 'package:fixnum/fixnum.dart';
+
+typedef void WebsocketChange();
+
+enum WebsocketMessage {
```
jigar-f, line 9: Can change `WebsocketMessage` to `_WebsocketMessageType` also let's add _, I don't think we want to export the enum to anywhere except this class.

jigar-f (approved):
Changes look good, Just added a few comments.

lib/common/ffi_subscriber.dart
```diff
@@ -3,12 +3,6 @@ import 'package:web_socket_channel/status.dart' as status;
...
 import 'common_desktop.dart';
 
-extension BoolParsing on String {
-  bool parseBool() {
-    return this.toLowerCase() == 'true';
-  }
-}
-
 class FfiValueNotifier<T> extends SubscribedNotifier<T?> {
   FfiValueNotifier(
```
jigar-f, line 7: I think we can remove most of the arguments now, Since we are using only `defaultValue` and `removeFromCache`

jigar-f:
I see there are a few places in sessionModel we are using `ffiValueBuilder`, I think we need migrated that as well

lib/common/websocket_subscriber.dart
atavism: Done

desktop/lib.go
atavism: Done

lib/common/ffi_subscriber.dart
atavism: Done

atavism:
> I see there are a few places in sessionModel we are using ffiValueBuilder, I think we need migrated that as well

I removed the remaining instances of it except for chat where we are just setting some default values

PR #1152, opened 2024-08-19
Open shepherd payment provider in system browser

atavism:
Resolves https://github.com/getlantern/engineering/issues/1545


atavism:
Tested on macOS

PR #1158, opened 2024-08-27
use new geolookup api

garmr-ulfr:
The geolocation, user info, and proxy config requests in `flashlight` have been merged into a single request. The new `flashlight/geolookup` now only stores the IP and country and `Refresh` has been removed. 

This PR relies on [flashlight PR](https://github.com/getlantern/flashlight/pull/1395) and, as such, is using that branch. Once the flashlight PR is merged, this will need to be updated to use the main branch again before this can be merged.


internalsdk/android_test.go
```diff
@@ -94,6 +93,10 @@ func TestProxying(t *testing.T) {
 		newResult, err := Start("testapp", "en_US", testSettings{}, testSession{})
 		require.NoError(t, err, "Should have been able to start lantern twice")
 		require.Equal(t, result.HTTPAddr, newResult.HTTPAddr, "2nd start should have resulted in the same address")
+
+		// give the client time to initialize. This helps with consistency in tests
+		time.Sleep(1 * time.Second)
```
anacrolix, line 98: Is there something you can expose to synchronize on that will guarantee that it's initialized? Sleeping in tests is always a bad idea and the tests are already super flakey.

internalsdk/android_test.go
```diff
@@ -160,7 +163,7 @@ func testProxiedRequest(helper *integrationtest.Helper, proxyAddr string, dnsGra
 
 	var buf []byte
 
-	buf, err = ioutil.ReadAll(res.Body)
+	buf, err = io.ReadAll(res.Body)
```
anacrolix, line 166: Can you add a sanity check to the length of the body here? Just check Response.ContentLength or whatever it's called, or add a io.LimitReader to some reasonably (large) value? Although I do see a timeout above, if it applies to body reading then it's moot.

garmr-ulfr: It thing it applies to the entire request, including reading the body.

internalsdk/android_test.go
garmr-ulfr: Removed. `Start` actually blocks until it's ready.

go.mod
```diff
@@ -1,14 +1,15 @@
...
-toolchain go1.22.8
+go 1.23
 
 // replace github.com/getlantern/flashlight/v7 => ../flashlight
+
 // replace github.com/getlantern/ipproxy => ../ipproxy
 // replace github.com/getlantern/fronted => ../fronted
 // replace github.com/getlantern/pathdb => ../pathDb/pathDb
 
+replace github.com/getlantern/flashlight/v7 => github.com/getlantern/flashlight/v7 v7.6.140-0.20241119222126-ff487ae5ee73
```
hwh33, line 11: We definitely don't want this.  This will make future attempts to update flashlight fail in surprising ways.  Could you update flashlight the normal way (remove this replace directive and just run `go get`)?

internalsdk/android_test.go
```diff
@@ -77,30 +78,48 @@ func (c testSession) IsPlayVersion() (bool, error)                       { retur
...
+func (c testSession) SetOnSuccess(enabled bool) {
+	if !enabled {
+		return
+	}
+	// we need to wait for flashlight.fastConnect dialer to "test" and get a successful connection
+	// before any dialer can be used ("no top dialer" error)
+	select {
+	case c.rdy <- true:
+	default:
+	}
```
hwh33, line 92: I'm a bit confused by this code given the comment.  This code doesn't wait at all.  Would you mind explaining?

internalsdk/replica.go
```diff
@@ -126,7 +125,7 @@ func (s *ReplicaServer) newHandler() (*replicaServer.HttpHandler, error) {
 			return s.Flashlight.FeatureOptions(config.FeatureReplica, options)
 		},
 		s.Session.GetCountryCode,
-		geolookup.Refresh)
+		func() {})
```
hwh33, line 128: Is making this a no-op okay @anacrolix?

internalsdk/android_test.go
garmr-ulfr: Added more to the comment to explain that this is only part of the solution and we would still need to actually listen for `c.rdy`.

internalsdk/android_test.go
```diff
@@ -77,30 +78,50 @@ func (c testSession) IsPlayVersion() (bool, error)                       { retur
...
+	if !enabled {
+		return
+	}
+	// we need to wait for flashlight.fastconnect dialer to "test" and get a successful connection
+	// before any dialer can be used ("no top dialer" error). fastConnect will call onSuccess when
+	// it gets a successful connection, this give us a way to listen for that before we continue.
+	select {
+	case c.rdy <- true:
+	default:
+	}
```
hwh33, line 93: Okay it took me a bit of investigation, but I think I understand now.  All you're doing is ensuring you wait for a successful connection before running the test.  I'd suggest the following to make this more clear to future readers:
- Do away with the flashlight-specific parts of this comment.
- Change the signal to closing of a channel.  That is the standard way of signaling readiness.  The code as-written is fragile in that it requires only a single waiting routine and one which is already waiting.  These requirements are confusing.

So this changes to:
```
func (c testSession) SetOnSuccess(enabled bool) {
	if !enabled {
		return
	}
	// Signal readiness on the first connection to a proxy.
	close(c.rdy)
}
```

And then in your test you have:
```
select {
case <-c.rdy:
case <-time.After(4 * time.Second):
    t.Fatal("session not ready in time")
}
```

And honestly the timeout part is optional since `go test` can specify and manage its own timeout (and the stack trace will tell you where the code got hung up).

garmr-ulfr: > Okay it took me a bit of investigation, but I think I understand now. All you're doing is ensuring you wait for a successful connection before running the test.

Sorry about that. I guess that wasn't as clear as I thought it was. 

> Change the signal to closing of a channel. That is the standard way of signaling readiness.

That's what I usually do but, in this case, the same `OnSuccess` callback is [called on every successful dial](https://github.com/getlantern/flashlight/blob/main/dialer/fastconnect.go#L90) regardless of the dialer used, so it will panic if `fastconnect` will test more than one. 

>  The code as-written is fragile in that it requires only a single waiting routine and one which is already waiting. These requirements are confusing.

Yeah, it requires `testSession.rdy` to be a buffered chan to avoid blocking `fastconnect`.

garmr-ulfr: And I just realized I should use `once.Do(close)`...

wendelcastro-ardanlabs (dismissed):

WendelHime (approved):

internalsdk/replica.go
garmr-ulfr: @anacrolix geolookup no longer supports forced refresh. The geolocation is now grouped in with the proxy config request.

PR #1167, opened 2024-09-09
Additional updates to ProClient

atavism:


internalsdk/webclient/defaultwebclient/rest.go
```diff
@@ -1,83 +0,0 @@
...
-
-func sanitizeResponseBody(data []byte) []byte {
-	var cleaned []byte
-	for _, b := range data {
-		if unicode.IsPrint(rune(b)) {
-			cleaned = append(cleaned, b)
-		}
-	}
-	return cleaned
-}
```
atavism, line 83: Just consolidating these changes below

internalsdk/auth/auth.go
```diff
@@ -50,25 +46,22 @@ type AuthClient interface {
...
-	}
-
-	wc := webclient.NewRESTClient(defaultwebclient.SendToURL(httpClient, baseURL,
-		func(client *resty.Client, req *http.Request) error {
-			prepareUserRequest(req, opts.UserConfig())
+func NewClient(baseURL string, userConfig func() common.UserConfig) AuthClient {
+	// The default http.RoundTripper is ChainedNonPersistent which proxies requests through chained servers
+	// and does not use keep alive connections. Since no root CA is specified, we do not need to check for an error.
+	// Temporarily disable use of ChainedNonPersistent
+	// rt, _ := proxied.ChainedNonPersistent("")
```
atavism, line 53: We can continue using a proxied.ParallelForIdempotent http.RoundTripper for the auth API. I tested it works fine, but I'm leaving the comment here in case we need to re-enable it.

PR #1168, opened 2024-09-10
Enable data cap on iOS

atavism:
IMG_6310

Resolves https://github.com/getlantern/engineering/issues/1458


atavism:
I did some testing and confirmed this is working. Make sure iOS throttling is turned on during testing or you may not receive any bandwidth updates

internalsdk/ios/ios.go
```diff
@@ -107,6 +108,10 @@ type StatsTracker interface {
 	UpdateStats(string, string, string, int, int, bool)
 }
 
+type BandwidthTracker interface {
```
jigar-f, line 111: thanks for making these changes.

ios/Runner/AppDelegate.swift
```diff
@@ -11,7 +11,7 @@ import UIKit
 
 // For IOS App Lunch time issue
 //https://developer.apple.com/videos/play/wwdc2019/423/?time=305
-@UIApplicationMain
+@main
```
jigar-f, line 14: Any reason why we are changing this ?

jigar-f (approved):
Changes LTGM, I'll test today and merge it.

jigar-f:
@atavism I think all API calls to user API fail due to `chained proxy unavailable` I feel this might be an issue on the client-side specific to IOS only. Seems like we are not calling [ListenAndServeHTTP](https://github.com/getlantern/flashlight/blob/1966e5f8cb92c1e5122b363ee4efb8a7ea8add98/flashlight.go#L496) that why it throwing error [here](https://github.com/getlantern/flashlight/blob/1966e5f8cb92c1e5122b363ee4efb8a7ea8add98/proxied/proxied.go#L544)

jigar-f:
This issue there because we changes our base URL for User API

PR #1174, opened 2024-09-12
Remove use of Android annotations and update gradle wrapper

atavism:


android/build.gradle
```diff
@@ -15,13 +15,11 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"
-        classpath 'com.android.tools.build:gradle:8.5.0'
+        classpath 'com.android.tools.build:gradle:8.7.0-beta01'
         classpath "com.google.protobuf:protobuf-gradle-plugin:0.9.1"
-        classpath 'com.google.gms:google-services:4.4.0'
+        classpath 'com.google.gms:google-services:4.4.2'
```
jigar-f, line 21: Why we are using `google-services` Can't we just remove it, Since we are not using firebase

android/build.gradle
```diff
@@ -15,13 +15,11 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"
-        classpath 'com.android.tools.build:gradle:8.5.0'
+        classpath 'com.android.tools.build:gradle:8.7.0-beta01'
```
jigar-f, line 19: @atavism Which Android studio version you are using? this is connected to the AS version Unless we get an incompatible error. Also why we are using beta version and not stable one?

.github/workflows/build-android.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       version: ${{ inputs.version }}
       version_file: ${{ inputs.version_file }}
       prefix: ${{ inputs.prefix }}
-    runs-on: macos-latest-xlarge
+    runs-on: ubuntu-latest
```
jigar-f, line 25: We are changing this coz somehow CI is not working on Android correct?

atavism: We are running out of disk space using a macOS runner

android/build.gradle
atavism: The beta one is the only one I could get working with Android Studio. Are you using the stable version?

jigar-f: Could you please try [Android Studio Jellyfish | 2023.3.1](https://developer.android.com/studio/archive) and keep gradle version same as 8.5.0

.github/workflows/build-android.yml
jigar-f: Thanks for making this change.

android/build.gradle
```diff
@@ -15,13 +15,10 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"
         classpath 'com.android.tools.build:gradle:8.5.0'
         classpath "com.google.protobuf:protobuf-gradle-plugin:0.9.1"
-        classpath 'com.google.gms:google-services:4.4.0'
```
jigar-f, line 22: @atavism I have removed google gms service

android/build.gradle
atavism: Yes doing that now

atavism: I'm happy to downgrade Android Studio, but just curious, does the latest version in the stable channel (Koala Feature Drop) not work for you? That's the version I've been using and it supports AGP 8.5-8.6

Screenshot 2024-09-16 at 4 27 47 AM

android/build.gradle
atavism: I think we just never removed it after we stopped using Firebase. Thanks for removing it

android/build.gradle
jigar-f: Sure, I can upgrade to Koala,

jigar-f:
@atavism Could you try now and see if the AGP 8.6.0 works?

PR #1180, opened 2024-09-18
Update Android to use flutter_inappwebview

atavism:


android/app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,7 @@
     <uses-permission android:name="com.android.vending.BILLING" />
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.READ_PRIVILEGED_PHONE_STATE" />
```
atavism, line 24: Fix for the following error:

```
  /Users/paul/.pub-cache/hosted/pub.dev/device_info_plus-10.1.2/android/src/main/kotlin/dev/fluttercommunity/plus/device_info/MethodCallHandlerImpl.kt:70: Error: Missing permissions required by Build.getSerial: android.permission.READ_PRIVILEGED_PHONE_STATE [MissingPermission]
                      Build.getSerial()
```

atavism:
I don't think we need WebViewActivity anymore with these changes, but the survey stuff is connected to it so I left it for now.

jigar-f (approved):
LGTM

jigar-f:
@atavism  The only thing is WebViewActivity was using a proxy setting and now in flutter, we are not using it.

atavism:
> @atavism The only thing is WebViewActivity was using a proxy setting and now in flutter, we are not using it.

Oh, great catch! Let me fix that.

atavism:
Ok, I'm using ProxyController setProxyOverride to configure the proxy the webview uses now too

PR #1181, opened 2024-09-20
Remove logs and fix webview issue.

jigar-f:


lib/common/ui/app_webview.dart
```diff
@@ -107,7 +107,7 @@ class AppBrowser extends InAppBrowser {
       navigationAction) async {
     final url = navigationAction.request.url!;
     if (url.scheme.startsWith("alipay")) {
-      await launchUrl(url, mode: LaunchMode.platformDefault);
```
jigar-f, line 110: You don't need to add await here, adding await makes an issue. launchUrl will to open that same URl and user will see that web page issue.

atavism (approved):

lib/common/ui/app_webview.dart
atavism: What issue are you experiencing? I just referenced how flutter_inappwebview calls launchUrl in the [example](https://github.com/pichillilorenzo/flutter_inappwebview/blob/4c58653113573558ea5be880aca931e85b23da7e/flutter_inappwebview/example/lib/in_app_webiew_example.screen.dart#L157) and [url_launcher](https://github.com/flutter/packages/blob/f54fe930753f6e44def213297bd8e794c846ad43/packages/url_launcher/url_launcher/lib/src/link.dart#L126).

PR #1185, opened 2024-09-24
delay call to setState with addPostFrameCallback

atavism:
For https://github.com/getlantern/engineering/issues/1652

```
══╡ EXCEPTION CAUGHT BY FOUNDATION LIBRARY ╞════════════════════════════════════════════════════════
The following assertion was thrown while dispatching notifications for ValueNotifier<bool>:
setState() or markNeedsBuild() called during build.
This VPNSwitch widget cannot be marked as needing to build because the framework is already in the
process of building widgets. A widget can be marked as needing to be built during the build phase
only if one of its ancestors is currently building. This exception is allowed because the framework
builds parent widgets before children, which means a dirty descendant will always be built.
Otherwise, the framework might not visit this widget during this build phase.
The widget on which setState() or markNeedsBuild() was called was:
  VPNSwitch
The widget which was currently being built when the offending call was made was:
  FfiValueBuilder<String>

When the exception was thrown, this was the stack:
#0      Element.markNeedsBuild.<anonymous closure> (package:flutter/src/widgets/framework.dart:5177:9)
#1      Element.markNeedsBuild (package:flutter/src/widgets/framework.dart:5189:6)
#2      State.setState (package:flutter/src/widgets/framework.dart:1223:15)
#3      _VPNSwitchState.build.<anonymous closure>.<anonymous closure> (package:lantern/vpn/vpn_switch.dart:118:15)
#4      _AdvancedSwitchState._handleControllerValueChanged (package:flutter_advanced_switch/flutter_advanced_switch.dart:279:23)
#5      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:437:24)
#6      ValueNotifier.value= (package:flutter/src/foundation/change_notifier.dart:559:5)
#7      _AdvancedSwitchState.didUpdateWidget (package:flutter_advanced_switch/flutter_advanced_switch.dart:108:24)
#8      StatefulElement.update (package:flutter/src/widgets/framework.dart:5789:55)
#9      Element.updateChild (package:flutter/src/widgets/framework.dart:3941:15)
#10     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5642:16)
#11     StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5780:11)
#12     Element.rebuild (package:flutter/src/widgets/framework.dart:5333:7)
#13     BuildScope._tryRebuild (package:flutter/src/widgets/framework.dart:2693:15)
[...]

PR #1186, opened 2024-09-25
Update system tray to use vpn status notifier

atavism:
Related to https://github.com/getlantern/lantern-client/pull/1185, this PR simplifies the system tray code and updates it to use the VPN notifier

I did some testing, and it's working well. ~The only issue I'm seeing is the VPN switch isn't being flipped on (but the status is changing) when I connect via the system tray menu~ (should be resolved by the latest commit)


atavism:
Hey @jigar-f, could you take a look at this one?

jigar-f (approved):
There is a small exception, while starting the lantern, I fixed it.

PR #1189, opened 2024-09-26
Added proxy setting screen.

jigar-f:
Ticket-:https://github.com/getlantern/engineering/issues/1408


lib/account/proxies_setting.dart
```diff
@@ -0,0 +1,49 @@
+import '../common/common.dart';
+
+@RoutePage(name: 'ProxiesSetting')
```
jigar-f, line 3: We can't not name  ProxySettings due webview has same class

jigar-f:
@atavism cloud you look into this one? Is there any work left on this ticket.

atavism:
Merged in the project structure changes.

Here's what the proxy settings page looks like on desktop:

Screenshot 2024-09-29 at 2 39 27 PM

I don't think we need to add anything else, right @Derekf5?

PR #1192, opened 2024-09-30
Added support for linux webview.

jigar-f:


jigar-f:
@atavism Do mind taking a look at this?

PR #1194, opened 2024-09-30
Update system proxy management to run FFI calls on isolate

atavism:
This PR updates the desktop FFI code for managing the system proxy to run on a separate isolate, improving overall stability. It resolves some issues we've been seeing with conflicting signal handling between Dart and Go by moving the FFI calls to a separate thread, which helps to avoid problematic interactions with the Go runtime signal stack

Working well for me on macOS


jigar-f:
@atavism I made few more changes, Do take look

jigar-f:
Also, there is one more issue, somehow bandwidth is not getting all data from desktop.

lib/core/service/lantern_ffi_service.dart
```diff
@@ -61,6 +68,57 @@ class LanternFFI {
...
+    if (!_isolateInitialized.isCompleted) {
+      await _initializeSystemProxyIsolate();
+    }
+
+    final responsePort = ReceivePort(); // Port to receive isolate's response
+    _proxySendPort?.send([vpnStatus, responsePort.sendPort]);
+
+    // Listen for the result (success or error)
+    final message = await responsePort.first;
+    if (message == "done") {
```
jigar-f, line 91: Before we are not checking error and done from isolate now we are doing it.

lib/features/vpn/vpn_switch.dart
```diff
@@ -49,42 +51,68 @@ class _VPNSwitchState extends State<VPNSwitch> {
         });
       });
     } else {
-      // This ui for desktop
-      return AdvancedSwitch(
```
jigar-f, line 53: I had to remove AdvancedSwitch and use new one, reason being there was issue with control.

lib/features/vpn/vpn_switch.dart
```diff
@@ -110,4 +138,16 @@ class _VPNSwitchState extends State<VPNSwitch> {
       );
     }
   }
+
+  Future<void> vpnProcessForDesktop(String vpnStatus) async {
+    final vpnNotifier = Provider.of<VPNChangeNotifier>(context, listen: false);
```
jigar-f, line 143: Now we are checking for errors, if errors are then we show the message and set the status.

desktop/lib.go
```diff
@@ -135,8 +135,13 @@ func saveUserSalt(salt []byte) {
 }
 
 //export sysProxyOn
-func sysProxyOn() {
-	go a.SysproxyOn()
+func sysProxyOn() *C.char {
+	err := a.SysproxyOn()
```
jigar-f, line 139: Now we are returning error on flutter end and showing error to user

lib/features/vpn/vpn_switch.dart
atavism: What issue are you experiencing with the control..? It works fine as is for me:

https://github.com/user-attachments/assets/88cafe37-2473-47f3-849f-bb4d84b71dbf

lib/core/service/lantern_ffi_service.dart
atavism: Just to give some context: I thought we could avoid some redundant response checking by sending the status update (if setting the system proxy succeeded or failed) via the websocket channel afterwards (actually, I don't think we need to do that anymore and can simplify things further with this setup).

Using a ReceivePort here to check if the isolate _received/processed_ the request though definitely makes sense. Thanks for making the change!

lib/features/vpn/vpn_switch.dart
atavism: I definitely like the change to use animated_toggle_switch now so I think the control issues are irrelevant.

atavism:
> Also, there is one more issue, somehow bandwidth is not getting all data from desktop.

Fixed.

PR #1198, opened 2024-10-02
Open webview in-app instead of a seperate window on macOS and Windows

atavism:
This PR updates the webview implementation to open the checkout page directly within the app on macOS and Windows, rather than in a separate window. This improves the user experience by keeping navigation seamless and consistent with the rest of the app.

It also includes changes to use [webview_windows](https://pub.dev/packages/webview_windows) and some fixes to open the FAQ, forum, and follow us pages in the system browser.

@Derekf5, I know you mentioned opening the webview in a window right next to the main interface. What do you think about opening it directly in-app instead? Here's a video of what it looks like.

https://github.com/user-attachments/assets/be59ba16-21f1-45ea-8b73-5ee94912c296


lib/core/app/app_webview.dart
```diff
@@ -1,7 +1,7 @@
 import 'package:flutter_inappwebview/flutter_inappwebview.dart';
-import 'package:flutter_windows_webview/flutter_windows_webview.dart';
 import 'package:lantern/core/utils/common.dart';
 import 'package:url_launcher/url_launcher.dart';
+import 'package:webview_windows/webview_windows.dart';
```
jigar-f, line 4: For Windows why can't we use flutter_inappwebview, which supports Windows already?

lib/core/app/app_webview.dart
```diff
@@ -87,79 +118,77 @@ class AppBrowser extends InAppBrowser {
...
 
-  static Future<void> openWindowsWebview(String url) async {
-    FlutterWindowsWebview().launchWebview(url);
-  }
+  // openWithSystemBrowser opens a URL in the browser
+  static Future<void> openWithSystemBrowser(String url) async =>
+      await InAppBrowser.openWithSystemBrowser(url: WebUri(url));
 
-  static Future<void> openWebview(String url) async {
+  static Future<void> openWebview(BuildContext context, String url) async {
```
jigar-f, line 177: There was an issue with opening web view in Linux, I fixed in #1192 , You might need to take pull

jigar-f (commented):
Just couple of comments

lib/core/app/app_webview.dart
atavism: I thought we had some issues with it on Windows, but I haven't tested it recently. Let me check if it works and if so happy to switch to it

Derekf5:
@atavism that is a nice improvement over my suggestion!   How does Shepherd look opened in the webview?

atavism:
@Derekf5 Here's what Shepherd looks like in the webview on Windows

https://github.com/user-attachments/assets/b517a4cb-a7a0-45ea-a093-4ed7419ba175

Derekf5:
perfect, just wanted to make sure the QR code fit in frame without scrolling, thanks @atavism!

lib/core/app/app_webview.dart
atavism: Ok we are using flutter_inappwebview on Windows now too

atavism:
I fixed the issue where the app opens with a Broken UI on Windows as well

image

lib/core/app/app_webview.dart
atavism: Thanks! I pulled in those changes

lib/core/app/app_webview.dart
```diff
@@ -30,17 +35,43 @@ class _AppWebViewState extends State<AppWebView> {
...
-        initialUrlRequest: URLRequest(url: WebUri(widget.url)),
-        initialSettings: settings,
-        onProgressChanged: (controller, progress) {
-          appLogger.i("Progress: $progress");
-        },
-      ),
+      showAppBar: false,
+      body: Platform.isWindows
+          ? _buildWindowsWebView(widget.url)
+          : InAppWebView(
```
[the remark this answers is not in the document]
atavism: I just re-enabled the app bar for now so the user has the option to go back

PR #1200, opened 2024-10-02
Use bandwidth checking same as other platforms.

jigar-f:


lib/features/vpn/vpn_bandwidth.dart
```diff
@@ -8,13 +8,8 @@ class VPNBandwidth extends StatelessWidget {
   Widget build(BuildContext context) {
     return sessionModel
         .bandwidth((BuildContext context, Bandwidth bandwidth, Widget? child) {
-      // User does not have bandwidth cap off
-      if (bandwidth.remaining > 0 || isDesktop()) {
```
jigar-f, line 12: There was an issue here, If use have not used any of the data here then it shows the progress bar as complete.

desktop/datacap/datacap.go
```diff
@@ -92,6 +99,27 @@ func (dc *dataCap) processQuota(out chan<- interface{}, quota *bandwidth.Quota)
 		})
 	}
 }
+func getBandwidth(quota *bandwidth.Quota) (int, int, int) {
```
jigar-f, line 102: @atavism I think should use what we are using on other platforms just for consistency and we can avoid some platform conditions on the flutter end.

jigar-f:
I am merging this, Let me know if you want me to revert this chanegs.

atavism:
OK, now that these changes were reverted, "When first installed, I saw a data cap, but when I did a clean install on another Windows user account, I saw no data cap. After about 10 minutes of use, the data cap showed up again." is an issue again.

jigar-f:
@atavism In this case, I think there might be some issue with the bandwidth check on the desktop, Could you see why there lag on that

PR #1208, opened 2024-10-09
Add mocks and create user after startup in Go

atavism:
Resolves https://github.com/getlantern/engineering/issues/1347 and https://github.com/getlantern/engineering/issues/1695


internalsdk/android.go
```diff
@@ -58,22 +57,33 @@ type Settings interface {
 	TimeoutMillis() int
 }
 
+// AdSettings is an interface for retrieving mobile ad settings from the global config
+type AdSettings interface {
```
jigar-f, line 61: @atavism Are we still using this anywhere? I didn't see it anywhere. If not can we just remove it?

android/app/src/main/kotlin/org/getlantern/lantern/service/LanternService.kt
```diff
@@ -33,8 +33,6 @@ open class LanternService : Service(), Runnable {
     }
 
     private var thread: Thread? = null
-    private val createUserHandler: Handler = Handler(Looper.getMainLooper())
-    private val createUserRunnable: CreateUser = CreateUser(this)
```
jigar-f, line 37: Thanks for making these changes, Looks much better now. 👍

internalsdk/common/ad.go
```diff
@@ -0,0 +1,15 @@
+package common
+
+// AdSettings is an interface for retrieving mobile ad settings from the global config
+type AdSettings interface {
```
jigar-f, line 4: Can't we use the same on `android.go`?

internalsdk/mocks/AdProvider.go
jigar-f, line 1: Can you please add some documentation on one of the files, How can we generate this mock class, and others you think might be helpful, that will help me as well

internalsdk/pro.go
jigar-f, line 1: Can't we move this code to utils.go

jigar-f (commented):
Just couple of comments.

internalsdk/common/ad.go
atavism: I moved it temporarily to common to avoid an import cycle in internalsdk with the mocks code, but that should be fixed now.

internalsdk/mocks/AdProvider.go
atavism: Added documentation to the README

internalsdk/pro.go
atavism: Done

PR #1210, opened 2024-10-09
Implement TrayContainer and refactor LanternApp to use it

atavism:
This PR introduces a TrayContainer, similar to the recent changes to add a WindowContainer, and updates LanternApp to conditionally wrap both containers only on desktop.

It removes direct usage of tray_manager from the home widget, which was previously responsible for initializing the tray.


jigar-f (approved):
LGTM

PR #1217, opened 2024-10-21
Address Crashes in 7.9.8 Beta Build

jigar-f:


android/app/src/main/kotlin/org/getlantern/lantern/service/LanternService.kt
```diff
@@ -42,12 +42,7 @@ open class LanternService : Service(), Runnable {
 
     override fun onCreate() {
         super.onCreate()
-        val serviceIcon: Int = if (LanternApp.session.chatEnabled()) {
-                    R.drawable.status_chat
-                } else {
-                    R.drawable.status_plain
-                }
-        helper = ServiceHelper(this, serviceIcon, R.string.ready_to_connect)
+        helper = ServiceHelper(this, null, R.string.ready_to_connect)
```
jigar-f, line 45: I already have fixed this issue on my previous PR, If null it will assign a value to serviceHelper class.

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
```diff
@@ -381,20 +393,26 @@ class SessionModel internal constructor(
...
                 val apps = buildJsonArray {
                     appsList.forEach { app ->
                         add(
                             buildJsonObject {
-                                val byte = ByteString.copyFrom(app.icon)
+                                val byte = app.icon!!
                                 put("packageName", app.packageName)
                                 put("name", app.name)
-                                put("icon", byte.toByteArray().toUByteArray().joinToString(", "))
+                                putJsonArray("icon") {
```
jigar-f, line 406: This will fix `OutOfMemeory` exception

lib/core/app/app_webview.dart
```diff
@@ -75,17 +75,25 @@ class AppBrowser extends InAppBrowser {
...
-          settings: ProxySettings(
-        proxyRules: [ProxyRule(url: "http://$proxyAddr")],
-        bypassRules: [],
-      ));
+    try {
+      var proxyAvailable = await WebViewFeature.isFeatureSupported(
+          WebViewFeature.PROXY_OVERRIDE);
+      if (proxyAvailable) {
+        ProxyController proxyController = ProxyController.instance();
+        final proxyAddr = await sessionModel.proxyAddr();
```
jigar-f, line 83: Is the proxy is empty do not set it, To avoid crash also added try-catch

myleshorton (approved):

android/app/src/main/kotlin/io/lantern/apps/AppsDataProvider.kt
```diff
@@ -11,7 +16,8 @@ class AppsDataProvider(
     private val applicationFilterPredicate: (ApplicationInfo) -> Boolean = { appInfo ->
         hasInternetPermission(appInfo.packageName) &&
                 isLaunchable(appInfo.packageName) &&
-                !isSelfApplication(appInfo.packageName)
+                !isSelfApplication(appInfo.packageName) &&
+                !isSystemApp(appInfo.packageName)
```
atavism, line 20: It looks like the filtering logic here is incorrectly excluding certain apps from the split tunneling list now. For instance, I'm no longer seeing apps that come pre-installed on devices (like Chrome, Gmail, etc).

android/app/src/main/kotlin/io/lantern/apps/AppsDataProvider.kt
```diff
@@ -41,6 +48,15 @@ class AppsDataProvider(
         return packageName == thisPackageName
     }
 
+    private fun isSystemApp(packageName: String): Boolean {
+        return try {
+            val appInfo = packageManager.getApplicationInfo(packageName, 0)
+            (appInfo.flags and ApplicationInfo.FLAG_SYSTEM) != 0
```
atavism, line 54: One way to resolve the issue is to exclude system apps that can be updated since this indicates they're user-facing

```kotlin
(appInfo.flags and ApplicationInfo.FLAG_SYSTEM) != 0 && (appInfo.flags and ApplicationInfo.FLAG_UPDATED_SYSTEM_APP) != 0)
```

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
atavism: Cool!

android/app/src/main/kotlin/org/getlantern/lantern/service/LanternService.kt
atavism: Sweet! Thanks for making this change. If you make defaultIcon optional with a default value of null, then you can just write

```kotlin
helper = ServiceHelper(this, R.string.ready_to_connect)
```

atavism:
LGTM in general. Just a couple comments after doing some testing.

android/app/src/main/kotlin/org/getlantern/lantern/service/LanternService.kt
jigar-f: Done

android/app/src/main/kotlin/io/lantern/apps/AppsDataProvider.kt
jigar-f: Thanks for noticing this, I think its best to remove this check, Since all OS have different kinds of system apps, which will complicate things.

jigar-f:
@atavism All the changes are done.

atavism:
LGTM!!

PR #1222, opened 2024-10-28
Another attempt to fix exceptions

jigar-f:
This PR fixes an issue with the OutOfMemoryError exception with a different approach. Now we are writing a file with the apps list and reading it on GO end. That should fix this issue 100%


android/app/src/main/kotlin/org/getlantern/lantern/plausible/PlausibleClient.kt
```diff
@@ -195,17 +195,22 @@ internal class NetworkFirstPlausibleClient(
...
-            Proxy(
-                Proxy.Type.HTTP,
-                InetSocketAddress(
-                    "127.0.0.1",
-                    uri.port,
-                ),
-            )
-        OkHttpClient.Builder().proxy(proxy).build()
+        if (hTTPAddr.isEmpty()) {
+            OkHttpClient()
```
jigar-f, line 199: If we don't have proxyAddr lets not crash and give it an empty client.

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
```diff
@@ -396,28 +401,53 @@ class SessionModel internal constructor(
...
+                val file = File(activity.cacheDir, "appsData.json")
+
+
+                // Write JSON data directly to the file to reduce memory usage
+                file.outputStream().use { outputStream ->
+                    Json.encodeToStream(buildJsonArray {
+                        appsList.forEach { app ->
+                            add(
+                                buildJsonObject {
+                                    val byte = app.icon ?: ByteArray(0)  // Safeguard for null icons
```
atavism, line 413: I think adding this safeguard is useful. What will the user see when an icon is missing? Blank space? Do we want to display a generic app icon as a placeholder?

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
```diff
@@ -396,28 +401,53 @@ class SessionModel internal constructor(
...
+                                        byte.toUByteArray().forEach { add(it.toInt()) }  // Store each UByte as an Int
+                                    }
                                 }
-                            }
-                        )
-                    }
+                            )
+                        }
+                    }, outputStream)
                 }
```
atavism, line 423: This looks like a great way to mitigate some of the memory issues! I think we could optimize further by implementing some caching here as well? For instance, you could serve the JSON file from cache if it already exists and add logic to refresh it only if it’s older than 24 hours. Additionally, you could minimize file writes by re-generating the file only if the installed apps differs from the last snapshot? We could always handle this in a subsequent PR but I figured I would mention it!

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
jigar-f: I don't think that is possible, Most of them have an app Icon, Anyway, I am going to add some placeholder on the Flutter end.

android/app/src/main/kotlin/io/lantern/model/SessionModel.kt
jigar-f: Thanks, Looking into it.

jigar-f: Changes are done.

jigar-f:
@atavism All changes are done, Do take look

jigar-f:
@atavism I think we should merge this and put out a new release.

atavism:
@jigar-f I'm testing the changes now.

atavism:
Working well for me

PR #1225, opened 2024-11-01
Use correct window size on Linux

atavism:


pubspec.yaml
```diff
@@ -100,7 +100,7 @@ dependencies:
   desktop_webview_window: ^0.2.3
 
   # Desktop
-  window_manager: ^0.4.2
+  window_manager: ^0.4.3
```
atavism, line 103: I noticed there's a new version of window_manager available so I'm updating it on this PR as well.

garmr-ulfr (approved):
LGTM! (Code and window size)

PR #1238, opened 2024-11-22
Update retry handler used by Pro Client

atavism:
This PR updates the retry handler for user creation, originally implemented for Android, to work across multiple platforms, including desktop


atavism:
@jigar-f I'm testing this now, but it would be great if you could start taking a look!

jigar-f:
@atavism I'll give it look tomorrow, thanks

desktop/auth.go
```diff
@@ -128,10 +128,6 @@ func logout() *C.char {
 	}
 
 	clearLocalUserData()
-	// Create new user
```
jigar-f, line 131: We need to create a new user when the user logout, Any reason why we are removing it?

jigar-f (commented):
I have created PR #1246 for this fix

desktop/lib.go
```diff
@@ -148,19 +148,23 @@ func setProxyAll(value *C.char) {
 //
 //export hasPlanUpdatedOrBuy
 func hasPlanUpdatedOrBuy() *C.char {
+	ctx := context.Background()
+	proClient := a.ProClient()
+	go proClient.PollUserData(ctx, a.Settings(), 10*time.Minute, proClient)
```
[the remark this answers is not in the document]
atavism: Yeah, I understand the concern about the polling duration, and the backend does handles it quickly, but this is intended to cover the duration of a full checkout flow (and with certain providers there's additional steps, like being redirected to the Alipay app). User checkout flows on average take longer than 2-3 minutes and with the exponential backoff, stopping after such a short amount of time would result in too few retries

desktop/auth.go
atavism: It would be great to move a lot of this logic out of the FFI methods to make them more lightweight and maintainable. Having a leaner FFI interface would also simplify the front-end integration and make it easier to switch between different core proxy libraries.. That said, I mainly overlooked adding these back!

internalsdk/android.go
```diff
@@ -683,7 +712,8 @@ func afterStart(wrappedSession Session, session PanickingSession) {
 
 	if session.GetUserID() == 0 {
 		ctx := context.Background()
-		go retryCreateUser(ctx, wrappedSession)
+		proClient := createProClient(wrappedSession, "android")
+		go proClient.RetryCreateUser(ctx, session, 10*time.Minute)
```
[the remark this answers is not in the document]
atavism: Thanks for catching this.. I didn't get a chance to test Android yet so I'll do that now and take a look at your updates!

desktop/auth.go
```diff
@@ -309,10 +305,5 @@ func deleteAccount(password *C.char) *C.char {
 	clearLocalUserData()
 	// Set user id and token to nil
 	a.Settings().SetUserIDAndToken(0, "")
-	// Create new user
-	// Create new user
-	if _, err := a.CreateUser(ctx); err != nil {
```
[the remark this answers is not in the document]
atavism: Ok I added back the calls to create a new user!

lib/core/service/websocket_subscriber.dart
```diff
@@ -88,12 +88,10 @@ class WebsocketSubscriber {
...
-            if (userLevel != null) {
-              if (userLevel == 'pro' || userStatus == 'active') {
-                sessionModel.proUserNotifier.value = true;
-              } else {
-                sessionModel.proUserNotifier.value = false;
-              }
+            final isLevelPro = userLevel != null && userLevel == 'pro';
+            final isStatusPro = userStatus != null && userStatus == 'active';
+            if (isLevelPro || isStatusPro) {
+              sessionModel.proUserNotifier.value = true;
```
jigar-f, line 94: We also need to make sure to set it to false.

desktop/app/pro.go
```diff
@@ -109,44 +107,44 @@ func (app *App) IsProUserFast(uc common.UserConfig) (isPro bool, statusKnown boo
 // It loops forever in 10 seconds interval until the user is fetched or
 // created, as it's fundamental for the UI to work.
 func (app *App) servePro(channel ws.UIChannel) error {
-	chFetch := make(chan bool)
-	ctx := context.Background()
-	go func() {
```
jigar-f, line 114: @atavism We were creating two users like in android, We don't need this anymore, Since we are creating user in afterStart

atavism: Thanks for removing this!

atavism:
> Just pushed some changes.

Thanks, looks great to me! Merging..

PR #1240, opened 2024-11-25
use go 1.22 again

atavism:


garmr-ulfr (approved):
LGTM. I really thought I would have had to change some things.

PR #1241, opened 2024-11-27
Show data cap after receiving first bandwidth update

atavism:
The data cap should show up right away now for free users (if it's enabled) with these changes


lib/features/vpn/vpn_bandwidth.dart
```diff
@@ -7,9 +7,13 @@ class VPNBandwidth extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return sessionModel
-        .bandwidth((BuildContext context, Bandwidth bandwidth, Widget? child) {
-          // User does not have bandwidth cap off
-      if (bandwidth.remaining > 0 ) {
+        .bandwidth((BuildContext context, Bandwidth? bandwidth, Widget? child) {
+      return sessionModel.proUser((context, isProUser, child) {
+        if (bandwidth == null || isProUser) {
```
atavism, line 12: This is the key change. I added an extra check here as well just to make sure we never show the data cap to Pro users

jigar-f:
Thanks, @atavism for making these changes, I'll test changes tomorrow.

jigar-f (approved):
LGTM

atavism:

PR #1245, opened 2024-12-02
Update hit_proxy scipt to work with new config format.

garmr-ulfr:
This PR updates the config filename and wraps the `dump-config` output with the keys flashlight expects. There might be a better way to handle this instead of hard-coding it.


WendelHime (approved):

atavism:
LGTM!

PR #1246, opened 2024-12-03
Fix user create api on android.

jigar-f:
This PR fixes the issue while calling user-create API on Android multiple times, Also sometimes `/plans` get called before the user creates still on going so other API calls fail because due user is not found.


atavism (approved):
LGTM!

PR #1250, opened 2024-12-04
Start config service sooner and don't copy mutex

myleshorton:
The current code was crashing for me when the config service was asked to process a config before it had started. There was also a warning about copying a mutex in a struct arg.


atavism (approved):

myleshorton:
Thanks @atavism!!

PR #1251, opened 2024-12-04
update http.RoundTripper used by Auth client

atavism:


internalsdk/auth/auth.go
```diff
@@ -51,35 +49,39 @@ type AuthClient interface {
...
+	frt := proxied.Fronted("auth_fronted_roundtrip", 10*time.Second)
 	rc := webclient.NewRESTClient(&webclient.Opts{
 		BaseURL: baseURL,
 		OnBeforeRequest: func(client *resty.Client, req *http.Request) error {
 			prepareUserRequest(req, userConfig())
 			return nil
 		},
-		HttpClient: httpClient,
+		HttpClient: &http.Client{
+			Transport: serialTransport{chained, frt},
```
atavism, line 80: This updates the auth client to use an http.Client that first attempts to connect via chained proxies and then falls back to using domain fronting with the custom name above (@jigar-f @hwh33 I'm not sure if we can just use `proxied.ChainedThenFronted()` here?).
Please let me know if this is correct otherwise feel free to change it to the one we should be using

jigar-f: I remember some work needs to be done on auth API access not sure if that was done or not. , Let's wait for @hwh33 on this.

myleshorton: Yeah I would just use `ChainedThenFronted`

myleshorton (approved):

PR #1252, opened 2024-12-05
Stickyconfig geo

garmr-ulfr:
This sets `country` in the generated config so the geolocation can be set. It will also check if running on linux and run the appropriate `make` cmd and not set device to macOS for flutter, so I don't have to manually change it every time I need to run it.


atavism:
@garmr-ulfr I think we still have the same issues building for Android with Go 1.23 https://github.com/getlantern/lantern-client/pull/1240

PR #1253, opened 2024-12-07
Use go 1.22 and downgrade some flashlight and fronted

jigar-f:
This PR fixes downgrade `Flashlight` and `Fronted` and other  due to complication issues on the Android


jigar-f:
cc @myleshorton We are downgrading fronted due to a complication issue on android.

garmr-ulfr (approved):

PR #1257, opened 2024-12-09
Use latest flashlight code that initializes domain fronting quicker

myleshorton:
This just uses the latest flashlight code that speeds up the selection of working domain fronts on startup, which also speeds up connecting to proxies.

PR #1259, opened 2024-12-10
Set Fronted on IOS

jigar-f:
The latest fronted  code crashed on IOS, the Reason being we were not setting fronted, on Android it's set by flashlight, This Pr should fix that issue

fronted.go

```
// Use a wrapper for fronted.NewDirect to avoid blocking
// `dualFetcher.RoundTrip` when fronted is not yet available, especially when
// the application is starting up
func (f frontedRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
	if f.opName != "" {
		op := ops.Begin(f.opName)
		defer op.End()
	}
	return fronter.RoundTrip(req)
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,8 +444,6 @@ set-version:
 	NEXT_BUILD=$$(($$CURRENT_BUILD + 1)); \
 	/usr/libexec/PlistBuddy -c "Set :CFBundleVersion $$NEXT_BUILD" $(INFO_PLIST)
 
-ios: macos build-framework ffigen
-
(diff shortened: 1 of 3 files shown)


myleshorton:
I think this should fix the iOS issue:

https://github.com/getlantern/lantern-client/pull/1257/files#diff-03a060d579a6fde0d9ee200008db0cc72d86f22b5f812e2ed4b6ef7eb3e7d90fR425

jigar-f:
Not needed anymore

PR #1267, opened 2024-12-16
use previous geolookup for ios

garmr-ulfr:
Since iOS uses it's own code to fetch proxy configs instead of the grouped request, it's unable to retrieve geolocation. This PR adds the old geolookup code exclusively for iOS to use. 

The `geolookup` package was copied directly from flashlight's old [geolookup package](https://github.com/getlantern/flashlight/blob/ebfaa2b5ed6208689f296d20312d5924c7bd3476/geolookup/geolookup.go).

### DIFF
--- a/internalsdk/ios/config.go
+++ b/internalsdk/ios/config.go
@@ -141,9 +141,8 @@ func (cf *configurer) Configure(userID int, proToken string, refreshProxies bool
 	if frontingErr := setupFronting(); frontingErr != nil {
 		log.Errorf("Unable to configure fronting, sticking with embedded configuration: %v", err)
 	} else {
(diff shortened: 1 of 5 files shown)


atavism:
I tried testing, but the country is still AF when I submit an issue report. I'm seeing this in the logs:

```
DEBUG ios.geolookup: geolookup.go:64 Could not get country: could not get geoinfo with timeout 1m0s: context deadline exceeded
```

internalsdk/ios/config.go
jigar-f, line 1: I had to call geolookup.Refresh() before GetCountry, now it works

jigar-f (approved):
Just one change added, otherwise LGTM

PR #1272, opened 2024-12-23
Deeplink Support for Desktop

jigar-f:


jigar-f:
The only way to test Windows is to create a build so I am merging these changes, I tested changes in macos working.
cc @atavism

PR #1275, opened 2024-12-25
Security assessment fixes

jigar-f:
This PR includes fixes for the issue from Leviathan.

Here is the ticket-: https://github.com/getlantern/engineering/issues/1741

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,8 +8,11 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.VIBRATE" />
-<!--    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />-->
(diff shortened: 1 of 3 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -8,8 +8,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.VIBRATE" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```
atavism, line 12: Do we need any of these permissions: READ_MEDIA_IMAGES, READ_MEDIA_VIDEO, or READ_MEDIA_AUDIO particularly for Replica?

atavism (approved):
LGTM, just one comment

android/app/src/main/AndroidManifest.xml
jigar-f: I don't think so, I tested replica downloading and uploading it seems to be working without that permission, Just for a safe bet, I added these permissions.

PR #1276, opened 2024-12-26
Desktop nits

jigar-f:
This PR includes

- Use mobile view in webview so the user does not have to scroll a lot.
- Fix exception on data cap on desktop.
- Add support for retry payment method if fails


lib/core/app/app_webview.dart
```diff
@@ -104,26 +104,26 @@ class _AppWebViewState extends State<AppWebView> {
...
+                displayZoomControls: false,
+                mediaPlaybackRequiresUserGesture: false,
+                allowsInlineMediaPlayback: false,
+                underPageBackgroundColor: Colors.white,
+                transparentBackground: true,
+                allowFileAccessFromFileURLs: true,
+                //We want to use mobile mode for webview on desktop
+                //  Since we are showign app on mobile size it will be better to use mobile mode
+                // showing recommended mode user has to scroll a lot
+                preferredContentMode: UserPreferredContentMode.MOBILE),
```
atavism, line 126: Thanks for fixing this

PR #1282, opened 2025-01-07
Sparrow SDK

jigar-f:


lib/core/service/survey_service.dart
```diff
@@ -0,0 +1,201 @@
...
+      domainName: "lantern.surveysparrow.com",
+      targetToken: AppSecret.UAEspotCheckTargetToken,
+      // Should Not Pass userDetails as const
+      userDetails: {});
+
+  final SpotCheck _myanmarSpotCheck = SpotCheck(
+      domainName: "lantern.surveysparrow.com",
+      targetToken: AppSecret.myanmarSpotCheckTargetToken,
+      // Should Not Pass userDetails as const
+      userDetails: {});
```
atavism, line 61: Maybe create a method to create SpotCheck instances dynamically based on country code to remove the need for all these separate variables?

lib/core/service/survey_service.dart
```diff
@@ -0,0 +1,201 @@
...
+        break;
+      case 'uae':
+        //UAE
+        _UAEspotCheck.trackScreen(screen.name);
+        break;
+      // This is just for testing
+      default:
+        _testingSpotCheck.trackScreen(screen.name);
+        break;
+    }
```
atavism, line 101: Same here.. you could add a utility method to fetch a SpotCheck for a given country code., and then call trackScreen on it

lib/core/service/survey_service.dart
```diff
@@ -0,0 +1,201 @@
...
+      // Should Not Pass userDetails as const
+      userDetails: {});
+
+  SurveyService() {
+    _createConfigIfNeeded();
+  }
+
+  void trackScreen(SurveyScreens screen) {
+    switch (sessionModel.country.value?.toLowerCase()) {
+      case 'ru':
```
atavism, line 69: Do we want to use enums for country codes to avoid hardcoding them in multiple places?

lib/core/service/survey_service.dart
```diff
@@ -0,0 +1,201 @@
...
+      appLogger.d('Failed to update vpnConnectCount: $e');
+    }
+  }
+
+  Future<bool> surveyAvailable() async {
+    try {
+      final filePath = await _surveyConfigPath;
+      final file = File(filePath);
+      if (await file.exists()) {
+        final content = await file.readAsString();
```
atavism, line 182: The logic for reading the config here could also be extracted into a reusable method

atavism:
Just a few comments on this one as well..

lib/core/service/survey_service.dart
jigar-f: If you are taking about `AppSecret.myanmarSpotCheckTargetToken` then that is not possible, this is a token needed need for each country.

lib/core/service/survey_service.dart
jigar-f: Good idea, thanks

lib/core/service/survey_service.dart
jigar-f: I agree, but I think if we just use shared_pref here, there might be less overhead for this, What do you think?

lib/core/service/survey_service.dart
jigar-f: Done

lib/core/service/survey_service.dart
jigar-f: Done

jigar-f:
Changes are done, Do you mind taking a look again?

PR #1284, opened 2025-01-09
8.0.4 Crash fixes

jigar-f:
This PR fixes top two crash of iOS.

- One with feature enable on Go side
- Other one with in app purchase module.


atavism (approved):
LGTM

PR #1287, opened 2025-01-13
Update Pre-release workflow.

jigar-f:
This PR updates the pre-release workflow to use platform-specific tags and also sends slack messages only if CI is successfully run


jay-418 (approved):
thanks!

PR #1288, opened 2025-01-14
Support App Open ads

jigar-f:


lib/core/service/injection_container.dart
```diff
@@ -1,10 +1,16 @@
...
 
 void initServices() {
   //Inject
-  sl.registerLazySingleton(() => AppPurchase());
-  sl<AppPurchase>().init();
+  if (isMobile()) {
+    sl.registerLazySingleton(() => AppPurchase());
+    sl<AppPurchase>().init();
+  }
+  sl.registerLazySingleton(() => AdsService());
```
atavism, line 15: Do we want to register the ads service on desktop or just mobile?

lib/core/ads/interstitial_ad_provider.dart
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  @override
+  Future<void> showAd() async {
+    if (isAdsShown) {
+      adsLogger.i("[Ads Manager] Google ad is already shown");
+      return;
+    }
+    _interstitialAd?.show();
```
atavism, line 76: Do we want to nullify _interstitialAd here after disposing it?

lib/core/ads/app_open_ads_provider.dart
```diff
@@ -0,0 +1,87 @@
...
+              onAdShowedFullScreenContent: (ad) {
+                isAdsShown = true;
+                adsLogger.i('[Ads Manager] Showing Ads');
+              },
+              onAdFailedToShowFullScreenContent: (ad, error) {
+                adsLogger.i(
+                    '[Ads Manager] onAdFailedToShowFullScreenContent callback');
+              },
+              onAdDismissedFullScreenContent: (ad) {
+                isAdsShown = true;
```
atavism, line 43: Should isAdsShown be set to false here after dismissal, to allow another ad to be shown? I'm not sure if we need to dispose the ad as well.

atavism:
Just a few comments, LGTM otherwise

lib/core/ads/app_open_ads_provider.dart
jigar-f: No, We need to show only one ads per session.

lib/core/service/injection_container.dart
jigar-f: Yes, because we are initializing `adHelper` at the top, we don't get an exception while running in desktop

lib/core/ads/interstitial_ad_provider.dart
jigar-f: Thanks for finding this.

jigar-f:
Changes are done, Do you mind taking a look again?

atavism:
LGTM!

PR #1291, opened 2025-01-22
Release CI Fixes.

jigar-f:
- Fix the issue with failing test flight step
- Support pushing out platform-specific versions such as android, ios , desktop

### DIFF
--- a/.github/workflows/prerelease-notification.yml
+++ b/.github/workflows/prerelease-notification.yml
@@ -2,10 +2,10 @@ name: Prerelease Notification
 
 on:
   workflow_run:
-#    workflows:
-#      - Publish releases
-#    types:
-#      - completed
+    workflows:
+      - Publish releases
+    types:
+      - completed
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/prerelease-notification.yml
```diff
@@ -12,6 +12,31 @@ jobs:
   create-release:
     runs-on: ubuntu-latest
     steps:
+      - name: Determine Platform
```
jigar-f, line 15: Now we are able to pass platform on slack message channel while CI pass

.github/workflows/release.yml
```diff
@@ -117,55 +113,66 @@ jobs:
       windows-arch: x64
 
   push-binaries:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     needs: [ determine-platform,set-version, build ]
     env:
       version: ${{ needs.set-version.outputs.version }}
       prefix: ${{ needs.set-version.outputs.prefix }}
+      platform: ${{ needs.determine-platform.outputs.platform }}
```
jigar-f, line 121: Now we push platform-specific tag then all builds will not make to lantern-repo. Just a platform we are pushing

atavism:
@jigar-f I tried to release 7.9.10 with these changes and ran into the following issue: https://github.com/getlantern/lantern-client/actions/runs/12916763218/job/36022795118

"This bundle is invalid. The value for key CFBundleShortVersionString [7.9.10] in the Info.plist file must contain a higher version than that of the previously approved version [8.0.5]"

PR #1292, opened 2025-01-22
Update ffigen config

atavism:


ffigen.yaml
```diff
@@ -4,7 +4,6 @@ description: Bindings to `liblantern.h`.
 output: 'lib/generated_bindings.dart'
 headers:
   entry-points:
-    - 'liblantern.h'
```
jigar-f, line 7: @atavism The reason I added this is because When we have great bindings for Windows and Linux we don't have arm64 or amd64 so binding does not generate if we are doing it on Windows and Linux, It will work only when we first run on mac.

PR #1293, opened 2025-01-22
Use latest flashlight with no proxied package and only kindling

myleshorton:
This eliminates all use of the proxied package in favor of the `getlantern/kindling` http client implementation.


desktop/autoupdate/autoupdate.go
```diff
@@ -28,18 +24,13 @@ var (
...
 )
 
 // Configure sets the CA certificate to pin for the TLS auto-update connection.
 func Configure(updateURL, updateCA string, iconURL func() string) {
 	setUpdateURL(updateURL)
 	fnIconURL = iconURL
-	httpClient.Store(
-		&http.Client{
-			Transport: proxied.ChainedThenFrontedWith(updateCA),
-		})
```
myleshorton, line 42: We are losing track of the update CA here -- have to think on that one

myleshorton: hmnn...actually I think this will break autoupdates as it stands

myleshorton: Ah interesting -- the autoupdate stuff actually handles this itself:

https://github.com/getlantern/autoupdate/blob/master/autoupdate.go#L122

myleshorton:
Merging!

PR #1296, opened 2025-01-24
Disable uploading build to Testflight for now

atavism:
I'm unable to push a new 8.0.0 release because of the following issue uploading the iOS build to Testflight: https://github.com/getlantern/lantern-client/actions/runs/12958781057/job/36150481991#step:11:16

> This bundle is invalid. The value for key CFBundleShortVersionString [8.0.0] in the Info.plist file must contain a higher version than that of the previously approved version [8.0.5]

I'm not sure how we fix this @jigar-f? I'm just going to disable uploading the iOS build to Testflight for now to get a release out. FYI @Derekf5

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,14 +175,14 @@ jobs:
         with:
           name: Lantern.ipa
 
-      - name: Upload Lantern to TestFlight
-        uses: apple-actions/upload-testflight-build@v1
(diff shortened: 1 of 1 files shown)


jigar-f:
@atavism Did you get chance to see this
https://github.com/getlantern/engineering/issues/1831#issuecomment-2606444163

atavism:
@jigar-f Just to clarify, if we want to release a new version to all platforms, we have to release it as 8.0.6, correct?

jigar-f:
Yeah, That is correct. not sure tho why need to push to all platforms.

PR #1297, opened 2025-01-27
Register custom scheme for windows.

jigar-f:
Looking into this more in https://github.com/getlantern/engineering/issues/1820
I found that without MSIX packaging, the HTTP/HTTPS protocols aren't supported on Windows. We need to migrate to using [[MSIX packaging](https://learn.microsoft.com/en-us/windows/uwp/launch-resume/web-to-app-linking#associate-your-app-and-website-with-a-json-file)] to support the HTTPS protocol.

In the meantime, I've added a custom scheme for Windows users. They can type this in any Windows browser:
```
lantern:/report-issue
lantern:/report-issue#beta
Lantern:/report-issue
Lantern:/report-issue#beta
```

cc @jay-418 @greySkiesMarsh


jigar-f:
I tested all the above links on my Windows all work.

jigar-f:
@atavism Do you mind taking a look at this?

jay-418:
@greySkiesMarsh I updated the [docs](https://github.com/getlantern/customer-support/wiki/Helpdesk#deep-linking) and [canned response](https://support.lantern.io/a/admin/canned_responses/folders/156000029743/responses/156000186957/edit).

Can you experiment with that and open a new ticket for me if it's wonky? I'm not sure how the link formatting needs to be done.

jigar-f:
@jay-418 It would be the same as we are doing in the other platforms, after `#` data will be added to the text box and without `#` it will be blank.

jay-418:
Thanks @jigar-f -- I should have been more clear that my concerns are about the formatting of the canned response in helpdesk which will deliver the link most often. I didn't test that end-to-end myself.

jigar-f:
That is actually the problem. Hyperlinks generally auto-detect on all platforms and most applications, but this kind of link, `lantern:/report-issue,` requires the user to copy it and put it into the browser, and then it will work.

jay-418:
Oof, thanks for clarifying @jigar-f. That's actually not much better than having to instruct users how to navigate within the app.

This might need MSIX to be truly useful for CS

lib/features/window/windows_protocol_registry.dart
```diff
@@ -0,0 +1,43 @@
...
+      String protocolCmdRegKey = 'shell\\open\\command';
+      RegistryValue protocolCmdRegValue = RegistryValue(
+        '',
+        RegistryValueType.string,
+        '"$appPath" "%1"',
+      );
+
+      final regKey = Registry.currentUser.createKey(protocolRegKey);
+      regKey.createValue(protocolRegValue);
+      regKey.createKey(protocolCmdRegKey).createValue(protocolCmdRegValue);
```
atavism, line 36: Should we close the registry keys after use here?

atavism:
@jigar-f Just one comment.. LGTM otherwise!

lib/features/window/windows_protocol_registry.dart
jigar-f: Done

jigar-f:
> Oof, thanks for clarifying @jigar-f. That's not much better than instructing users how to navigate within the app.
> 
> This might need MSIX to be truly useful for CS

@jay-418 Actually we can write a script on the lantern.io/report-issue page for windows, that does the work of opening the app.

PR #1302, opened 2025-02-05
Android prod updates

jigar-f:
- Get Rid of the sentry gradle plugin, It Seems like it was not uploading the debug symbol in sentry, it was just uploading mapping files.
- Remove panics from internal sdk.


atavism (approved):
LGTM!

PR #1303, opened 2025-02-07
Added check points for uae and iran.

jigar-f:


myleshorton:
Added a few tweaks, but I think this should be basically ready @atavism

jigar-f:
@myleshorton I tried your DNS server pins Multiple pins are coming as not reachable, I just tried with [pine.pe](https://ping.pe/140.117.167.174), Any idea why?
Screenshot 2025-02-10 at 11 36 33 AM

myleshorton:
I think we should just be using this:

https://pub.dev/packages/connectivity_plus

jigar-f:
@myleshorton the problem with using basic connectivity is it does not give accurate results, The User might have a Wi-Fi connection but their internet might not work, The User might have mobile data enabled but they don't have Data, So at some point, we might have to ping some data to get better results.

myleshorton:
True, but they are also going to know that generally. I'm not strictly opposed to this, but we should be very careful with it, as it's causing more problems than it solved currently.

It does still use connectivity plus under the hood, right? Like it only does checks when you should be connected?

jigar-f:
@myleshorton  Yeah, that's correct. What if I switch back my DNS server and put out the beta just for IOS and see how that works?
I have a storage feeling that in UAE I think Google is not blocked also in Iran, they unblocked the Google Play Store and so to me, It seems like it will work, and for China and Russia, we already have robust sites to check.

jigar-f:
@myleshorton  Do you see any security issue with that, If so I would remove it now...

myleshorton:
We shouldn't use google domains for checking anything. They might be unblocked in some places now, but we should assume it's always blocked, as it usually is.

myleshorton:
I already swapped the DNS server that was giving problems.

myleshorton:
This also include the important fix to not consider the user disconnected on any single DNS failure

jigar-f:
yeah, I got that change, thanks for making it.

PR #1304, opened 2025-02-07
Android DB updates

atavism:
This includes some updates to..

- properly quit the async handler thread when models are destroyed
- additional nil checks
- avoid redundant updates when checking split tunneling


internalsdk/session_model.go
```diff
@@ -829,15 +829,11 @@ func checkSplitTunneling(m *SessionModel) error {
...
-			return pathdb.Put(tx, pathSplitTunneling, false, "")
-		})
+		m.setSplitTunneling(false)
+		return err
 	}
 	if !tunneling {
-		log.Debugf("Split Tunneling value is %v", tunneling)
-		return pathdb.Mutate(m.db, func(tx pathdb.TX) error {
-			return pathdb.Put(tx, pathSplitTunneling, false, "")
-		})
```
atavism, line 840: If the split tunneling value is already false, we don't need to write it again

android/app/src/main/kotlin/io/lantern/model/GoModel.kt
```diff
@@ -29,36 +29,37 @@ open class GoModel<M : internalsdk.Model>(
     override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
         activeSink.set(events)
         val args = arguments as Map<String, Any>
-        val subscriberID = args["subscriberID"] as String
-        val path = args["path"] as String
-        val details = args["details"]?.let { it as Boolean } ?: false
+        val subscriberID = args["subscriberID"] as? String ?: return  // Exit early if null
+        val path = args["path"] as? String ?: return
```
atavism, line 33: Just added some additional checks to exit early here if either the subscriberID or path are null

android/app/src/main/kotlin/io/lantern/model/BaseModel.kt
```diff
@@ -258,11 +258,17 @@ abstract class BaseModel(
...
+            destroy()
+        }
     }
 
     fun destroy() {
         activeSubscribers.forEach {
             db.unsubscribe(it)
         }
+        // Quit the async handler thread
+        asyncHandlerThread.quitSafely()
```
atavism, line 273: This just makes sure the background thread is shut down gracefully when the model is destroyed

myleshorton (approved):

android/app/src/main/kotlin/io/lantern/model/GoModel.kt
jigar-f: Sounds good

internalsdk/session_model.go
jigar-f: Got it

jigar-f (approved):
LGTM

PR #1307, opened 2025-02-11
Build tidy

anacrolix:
Some improvements to Makefile and README while testing replica update builds on Android and MacOS.


atavism:
Thanks a lot for making these improvements, @anacrolix! This one also LGTM

PR #1308, opened 2025-02-11
Replica update

anacrolix:
Updates to latest replica. I had to pull in Go 1.23 because it's used in dependencies, but I started the process of testing that and ensured it builds on CI.


jigar-f:
@anacrolix When I tried to upgrade to the latest go there were some issues, on gomobile, It was crashing on Android. Not sure where that ticket is, but I'll find it.

jigar-f:
Here is the ticket-: https://github.com/getlantern/engineering/issues/1789#issuecomment-2638019109

anacrolix:
Looks like we can use go 1.23.6. Let me set it to that and retry?

atavism:
@anacrolix I just tested with Go 1.24.0 and the app is running fine (no crashes). Happy to merge this unless you want to make some additional changes!

PR #1315, opened 2025-02-13
Fix binding issue.

jigar-f:
This PR fixes the issue where the binding was generated on the Build folder, which is why FFI header bindings were not reflecting properly.


atavism:
Can you take a look at https://github.com/getlantern/lantern-client/pull/1316? We shouldn't have the generated headers committed anyway

PR #1316, opened 2025-02-13
Copy headers and update ffigen.yaml

atavism:


atavism:
OK CI successfully builds on all supported platforms https://github.com/getlantern/lantern-client/actions/runs/13317096491

PR #1320, opened 2025-03-11
Fix incorrect Windows installer name in production CI

atavism:
This updates CI to copy the Windows installer to the correct location as lantern-installer.exe, removing the unnecessary x64 extension (which I think comes from lantern-desktop)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -327,7 +327,7 @@ jobs:
         if: matrix.platform == 'windows'
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


myleshorton:
Oh wait sorry I know the 32 bit installer is compatible with everything (works on 64 bit) but the 64 bit will break on 32 bit right?

atavism:
@myleshorton Thanks! There isn't a 32-bit installer anymore: Flutter only supports 64-bit Windows

myleshorton:
Ah got it! Thanks

PR #1321, opened 2025-03-12
Fix issues linking device via email on desktop

atavism:
Resolves https://github.com/getlantern/engineering/issues/1991

image

### DIFF
--- a/lib/features/device_linking/authorize_device_via_email_pin.dart
+++ b/lib/features/device_linking/authorize_device_via_email_pin.dart
@@ -72,9 +72,7 @@ class AuthorizeDeviceViaEmailPin extends StatelessWidget {
   Future<void> onDone(String code, BuildContext context) async {
     try {
       context.loaderOverlay.show();
-      await sessionModel.validateDeviceRecoveryCode(code,email);
-    //once code is validated, approve the device
-      await sessionModel.approveDevice(code);
+      await sessionModel.validateDeviceRecoveryCode(code, email);
       pinCodeController.clear();
       context.loaderOverlay.hide();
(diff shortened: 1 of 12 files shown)


lib/features/device_linking/authorize_device_via_email_pin.dart
```diff
@@ -72,9 +72,7 @@ class AuthorizeDeviceViaEmailPin extends StatelessWidget {
   Future<void> onDone(String code, BuildContext context) async {
     try {
       context.loaderOverlay.show();
-      await sessionModel.validateDeviceRecoveryCode(code,email);
-    //once code is validated, approve the device
-      await sessionModel.approveDevice(code);
+      await sessionModel.validateDeviceRecoveryCode(code, email);
```
jigar-f, line 75: What is after validate code? How does this device get added to pro?

myleshorton:
Can we pull this in?

jigar-f:
I am not sure,  How this device gets added to pro. We are not calling API after validate

Derekf5:
Hey @atavism @jigar-f are we able to pull this in?

lib/features/device_linking/authorize_device_via_email_pin.dart
atavism: The device is added if the code is valid and call to `/user-link-validate` is successful: https://github.com/getlantern/lantern-cloud/blob/main/cmd/api/pro-server/store/link.go#L47

atavism:
I cleared devices, requested a new link code via email, validated it, confirmed the device is added and the UI transitions to Pro.

PR #1345, opened 2025-06-13
Update to version of flashlight that uses fronted that opens fewer fi…

myleshorton:
This is for https://github.com/getlantern/engineering/issues/2239


myleshorton:
Pulling this in despite test failures as it's clearly unrelated.

PR #1363, opened 2025-07-09
Update how we handle redirect URLs on checkout

atavism:


lib/features/checkout/utils.dart
```diff
@@ -0,0 +1,26 @@
...
+  if (url.startsWith('http://')) {
+    if (url.contains("alipay")) {
+      url = securePaymentRedirectUrl(url);
+    } else {
+      await openWithSystemBrowser(url);
+      return;
+    }
+  }
+  await openWebview(context, url, title);
+}
```
atavism, line 26: Tested on macOS. this solution seems to work well across payment providers

myleshorton (approved):

pubspec.yaml
```diff
@@ -126,7 +126,7 @@ dependencies:
   in_app_purchase: ^3.2.1
 
   # Ads
-  google_mobile_ads: ^6.0.0
+  google_mobile_ads: ^5.3.1
```
atavism, line 129: I had to downgrade a few dependencies to build Android and preserve API Level 21 support for now. Will fix this in the next release

PR #1364, opened 2025-07-10
IOS Consent

jigar-f:


lib/common/ui/custom/dialog.dart
```diff
@@ -187,28 +187,23 @@ class CDialog extends StatefulWidget {
...
+    CDialog(
+      title: 'What Info We Collect',
+      description: "collect_info_message".i18n,
+      agreeText: 'accept_and_continue'.i18n,
+      dismissText: 'decline'.i18n,
+      agreeAction: () async {
+        context.maybePop();
+        onConsent.call();
+        return false;
+      },
```
atavism, line 205: Oh, I did have one quick comment: do we want to show the user a message if they tap Decline? Like 'You must accept to continue', or just close the dialog? Otherwise, LGTM!

PR #1365, opened 2025-07-18
update issue map key for cannot link devices

atavism:


jay-418 (approved):
thanks for the quick turnaround!

PR #1384, opened 2026-07-24
Fix: Windows crash when clicking tray "Show" menu item

teamlantern:
## Summary
Auto-fix for Freshdesk ticket [#180333](https://lantern.freshdesk.com/a/tickets/180333).

**Root cause:** On Windows, clicking the tray icon's "Show" menu item called `windowManager.setSkipTaskbar(false)`, which reaches `WindowManager::SetSkipTaskbar` in window_manager 0.5.1 (`windows/window_manager.cpp:958`). That method dereferences the `ITaskbarList3* taskbar_` member, which is only initialized by `CoCreateInstance` inside `WindowManager::WaitUntilReadyToShow()`. Lantern never calls `windowManager.waitUntilReadyToShow()`, so `taskbar_` is null for the process lifetime and the call crashes the process with an access violation. The affected user (Windows, China, 8.3.7) saw 11 silent process deaths in 31 minutes and could never open the Lantern home page.

**What changed:** Removed the `windowManager.setSkipTaskbar(false)` call from the tray "Show" handler in `lib/features/tray/tray_container.dart`. The call is unnecessary — the window is never taskbar-skipped on Windows (`setSkipTaskbar(true)` is macOS-only, in `lib/features/window/window_container.dart`), so this is a pure crash-removal with no behavior change. The window is still shown/focused via `windowManager.show()` on tray mouse-down and `windowManager.focus()` in the handler.

## Test plan
- [ ] On Windows: launch, close the main window to tray, right-click tray icon → "Show" — window restores without the process dying
- [ ] Check for regressions in related code paths (tray menu, window show/hide on macOS/Linux)

---
*Auto-generated by `/ticket-fix`.*


jay-418:
Nice try @teamlantern, but this is the wrong repo.

